PR #182, opened 2020-06-09
updated for wop 1.9

christophe-david:
A fix was needed for compatibility with wop v1.9.

The missing depth argument has been added.

Besides, wop 1.9 does not crash anymore if `xfoil` option is True (was crashing because of the connection of non promoted variables). Thank you @relf 😉 

v1.9 is now set as requirement for FAST-OAD.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -64,22 +64,22 @@ jobs:
       - name: Unit tests
         run: |
           poetry run pytest src
-          poetry run coverage xml  # for Codacy coverage
+          poetry run coverage xml  # for sending coverage report
         shell: bash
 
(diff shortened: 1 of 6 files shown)


relf:
I suggest to keep a default depth to 3. It gives the high level structure of the code where XDSMs are somehow most valuable while the generation run more quickly and the result is a bit less overwhelming for the user as the XDSMs are less numerous.

christophe-david:
Actually, I have even set the default depth to 2. It is the depth level I think most suitable for our current process structure.

Deeper levels are still dark places where new users should not wander around 😄

relf:
You're right 2 is better! I made a typical off-by-one error and wop default should be 2 too. :smile:

PR #202, opened 2020-09-01
Fix openmdao

christophe-david:
This PR fixes the compatibility with OpenMDAO 3.2.

OpenMDAO 3.2 automatically creates, during setup(), an IndepVarComp instance called `_auto_ivc` that feeds all the unconnected inputs. The creation of these variables relies on the definition of defaults in the calls to `add_inputs`, but if linked variables are declared differently (different value or different units), a call to `set_input_defaults()` is needed to avoid a fatal error message.

For FAST-OAD, it means that:
- detection of unconnected inputs, that should be in input file, has to take into account the presence of the `_auto_ivc`
- all components that use a same variable should be consistent about their units (and its value, but it is generally `np.nan`), which is a pity, because it prevents from using the unit conversion capability of OpenMDAO.
- for variables where we want to use the unit conversion capability of OpenMDAO, usage of `set_input_defaults()` is needed, and it is done in a "patch" in `fastoad/models/defaults.py`. It helps on the 2 variables that FAST-OAD models declare with different units, but custom models that will use other existing variables will have to use the already declared units.


src/fastoad/models/defaults.py
```diff
@@ -0,0 +1,58 @@
...
+    :param model: the model where defaults will be added in place.
+    """
+
+    if version.parse(openmdao.__version__) < version.parse("3.2"):
+        return
+
+    variables = {
+        "data:geometry:wing:sweep_25": dict(val=np.nan, units="deg"),
+        "data:TLAR:range": dict(val=np.nan, units="NM"),
+    }
```
ScottDelbecq, line 37: Ajouter un TODO, on sait jamais on trouvera peut être une solution plus générique une jour

ScottDelbecq (approved):

src/fastoad/models/defaults.py
christophe-david: Fait en 2b312320d29dafbe612ee88398bc722923500aab (et j'ai plutôt mis un FIXME)

christophe-david: Ok, pas réveillé. Fait en e43cb93fb58bc1d9096f59e796fe4fc6c646e411

PR #231, opened 2020-10-08
Compatibility openmdao3.4

### LINKED ISSUE #230: Incompatibility with OpenMDAO 3.4
FAST-OAD crashes if used with OpenMDAO 3.4.

A correct usage of their new method `get_io_metadata()` should help avoiding such incompatibility in the future.

christophe-david:
This PR makes FAST-OAD compatible with OpenMDAO 3.4. Resolves #230.

Their is now no more usage of OpenMDAO private API, but used public API is not compatible with OpenMDAO <3.4.

Dependencies have been updated to enforce version 3.4+ of OpenMDAO (and version 1.12+ of wop, also for compatibility with OpenMDAO 3.4).

We also make good use of the new feature that allow variable shapes to be decided from connections. It allows to shape all polar-related variables without having to propagate a constant through all modules.


christophe-david:
> bad-continuation duplicate!?

Arf, zut 😞 
Dis donc, t'es attentif 😄

src/fastoad/openmdao/variables.py
```diff
@@ -472,15 +493,43 @@ def from_problem(
...
-        prom2abs.update(model._var_allprocs_prom2abs_list["output"])
-
-        for prom_name, abs_names in prom2abs.items():
-            if not promoted_only or "." not in prom_name:
-                # Pick the first
-                abs_name = abs_names[0]
-                metadata = deepcopy(model._var_abs2meta[abs_name])
+        # prom2abs = {}
+        # prom2abs.update(model._var_allprocs_prom2abs_list["input"])
+        # prom2abs.update(model._var_allprocs_prom2abs_list["output"])
```
ScottDelbecq, line 498: Quelques commentaires qui traînent.

ScottDelbecq (approved):
RAS juste quelques commentaires qui traînent dans le code (cf mon com n°2).

src/fastoad/openmdao/variables.py
christophe-david: Merci.
Corrigé en e67b8a275070d4c6b66c6a0d4fc863dfddb40557

PR #233, opened 2020-10-13
Main landing gear

ScottDelbecq:
This PR proposes minor changes:

- the computation time is now estimated using time() instead of process_time as this last method was providing under-estimated computation times when compared to user-side time
- the estimation of the main landing gear center of gravity now uses an ExplicitComponent instead of an ImplicitComponent + Group + Newton solver


christophe-david (approved):
C'est good

PR #236, opened 2020-10-13
Changed units of data:weight:aircraft_empty:mass from "m" to "kg".

### LINKED ISSUE #229: Wrong unit for "data:weight:aircraft_empty:mass"
Hello,

The unit of "data:weight:aircraft_empty:mass" is meter instead of kg in both the input file and the ComputeCGLoadCase classes.

Vincenzo

christophe-david:
Fixes #229


ScottDelbecq (approved):
RAS !

PR #237, opened 2020-10-14
Fix compressibility drag

christophe-david:
This PR fixes computation of compressibility drag, because only half of the model from SUPAERO course was implemented, resulting in complete negation of effect of sweep angle and thickness ratio on the compressibility drag.

But fixing the formula led to increasing the compressibility drag for CeRAS case. To keep current results (until we go through a thorough validation process), the change is balanced by reducing induced drag with the coefficient for winglet effect.

Nevertheless, even with this fix, compressibility drag of modern aircraft may still be overestimated by the model, so the variable `tuning:aerodynamics:aircraft:cruise:CD:compressibility:ceiling` has been added to cap the model result. Even with the default, very large, value, it allows to give better looking polar plots (with a strange break in the curve, indeed, but at least CD values do not go exponential any more).

### DIFF
--- a/src/fastoad/models/aerodynamics/components/cd_compressibility.py
+++ b/src/fastoad/models/aerodynamics/components/cd_compressibility.py
@@ -22,10 +22,18 @@ from openmdao.core.explicitcomponent import ExplicitComponent
 
 
 class CdCompressibility(ExplicitComponent):
+    """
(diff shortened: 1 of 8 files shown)


ScottDelbecq (approved):
RAS. Je pense que c'est une bonnne solution à court terme d'utiliser un facteur correctif.

PR #240, opened 2020-10-30
Devops fixes

christophe-david:
This PR brings 2 fixes:
- Black pre-commit hook now only requires Python 3 instead of specifying Python 3.7
- docs/requirements.txt is now generated without hashes, which avoids compilation failures in ReadTheDocs


ScottDelbecq (approved):
ça à l'air de mieux se passer effectivement !

PR #254, opened 2020-12-11
Now api.generate_inputs() returns the path of generated file.

### LINKED ISSUE #253: api.generate_inputs() should return the path of generated input file
Then for instance, we could simply have in a notebook:
```python
input_file_path = api.generate_input_file("my_conf.toml")
api.variable_viewer(input_file_path)
```

christophe-david:
Closes #253


ScottDelbecq (approved):
Très bonne idée !

PR #259, opened 2021-01-22
Change in module declaration

christophe-david:
`OpenMDAOSystemRegistry` has been deprecated and replaced by `RegisterOpenMDAOSystem` decorator that shares same parent class with `RegisterPropulsion` decorator.

Documentation has been [updated](https://fast-oad.readthedocs.io/en/module-declaration/documentation/custom_modules/add_modules.html#register-your-system-s).

Also, a minor change in documentation is to set TOC depth to 1 in RTD home page, to avoid having the big change log at first sight.


src/fastoad/models/performances/breguet/tests/test_breguet_openmdao.py
```diff
@@ -84,6 +85,7 @@ def test_breguet():
 
 
 def test_breguet_with_rubber_engine():
+    print(BundleLoader().get_factory_names())
```
ScottDelbecq, line 88: print qui traine

ScottDelbecq (approved):
Un commentaire mineur. En tout cas ça fonctionne chez moi !

src/fastoad/models/performances/breguet/tests/test_breguet_openmdao.py
christophe-david: Bien vu :)
Corrigé en e7e35c6207467921f3baba34c9aa03dcc83a2414

PR #274, opened 2021-03-10
Issue 269 dependencies check

### LINKED ISSUE #269: Check thoroughly dependencies requirements (minimal versions)
Minimum versions should be set for all main dependencies of the project, to ensure users won't get incompatible installations.

ScottDelbecq:
This PR aims at providing the minimal versions of dependencies for FAST-OAD to work on all platforms and for python 3.7 and later versions.

The python 3.7 release date (28 June 2018) has been chosen as the reference date for finding the minimal versions based on their release date history. Then, minimal versions of some packages (e.g. scipy, numpy,...) have been upgraded until tests and notebooks were running properly.

Closes #269.


christophe-david:
Thank you for this important work.

I tried and test the minimum package versions with a Python 3.8.8 and I had to ask for lxml 4.4.1 and matplotlib 3.1.2 to get Python 3.8-compatible wheel packages that allowed a correct installation.

I don't think your current specifications would cause any problem, because a Python 3.8 user will naturally have compatible package versions, I guess, and we do not force any version that would be incompatible with Python 3.8.
But Poetry allows to handle that gracefully, as can be seen [here](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies) (text provided below). And doing so, we won't have to retrieve the information about Python 3.8 compatibility when we will abandon 3.7 (in a long time, though).

## from Poetry documentation

### Multiple constraints dependencies

Sometimes, one of your dependency may have different version ranges depending on the target Python versions.

Let's say you have a dependency on the package foo which is only compatible with Python <3.0 up to version 1.9 and compatible with Python 3.4+ from version 2.0: you would declare it like so:

```toml
[tool.poetry.dependencies]
foo = [
    {version = "<=1.9", python = "^2.7"},
    {version = "^2.0", python = "^3.4"}
]
```

pyproject.toml
```diff
@@ -43,23 +43,34 @@ classifiers = [
...
+    {version = "^3.0.0", python = "^3.7"},
+    {version = "^3.1.2", python = "^3.8"}
+]
+plotly = "^4.5.0"
+ipywidgets = "^7.3.0"
+ipysheet = "^0.3.0"
+tomlkit = "^0.5.3"
 wop = "^1.12.0"
-strictyaml = "1.1.0"
+strictyaml = "^1.1.0"
```
christophe-david, line 71: Agreed.

pyproject.toml
```diff
@@ -43,23 +43,34 @@ classifiers = [
...
-pandas = "^1"
+numpy = [
+    {version = "^1.16.0", python = "^3.7"},
+    {version = "^1.18.0", python = "^3.8"}
+]
+scipy = [
+    {version = "^1.3.0", python = "^3.7"},
+    {version = "^1.4.0", python = "^3.8"}
+]
+pandas = "^1.1.0"
```
christophe-david, line 54: I had forgotten we need such a recent version of pandas.
It convinces me even more that we should not bother with too old versions for other packages.

christophe-david (changes requested):
Thanks for the update.
Unfortunately, you will hate me, because I changed my mind again 😅.

I tried to test the minimum versions, and I got (strange) problems on macOS with `matplotlib` 3.0.0 with Python 3.7. I had to go up to 3.1.0 to have successful tests, which is very close to the minimum requirement for Python 3.8.
Considering that, considering also that we require a much recent `pandas`, and finally considering that Python 3.7 is now an old version, I think we should simply require versions that are compatible with Python 3.8.

I applied this logic in the branch `issue-269_dependencies_check_cd`. If you agree with it, you may simply get the 2 commits in your branch.

(BTW, when doing the final `poetry update`, I noticed that OpenMDAO 3.8 is out!)

ScottDelbecq:
Haha no worries! This is what PR are made for... Thanks for the commits. I think we are ready for a merge!

Issue #271, opened 2021-03-02
Validation of results on CeRAS


christophe-david:
Creating a notebook that shows how to replay CeRAS-01 results

PR #275, opened 2021-03-11
CeRAS case study #271 

Martin-Delavenne:
This PR proposes a new case study notebook to retrieve CeRAS design characteristics from FAST OAD process. It also provides exemple a operational mission usage.


christophe-david (commented):
Thank you for this good work.

Well, this review is a bit particular.

Since your branch was based, as we agreed, on an intermediate version `operational-mission` branch, that is now merged into master branch, I did the following operations:
1. moved your branch `CeRAS-tutorial` into `CeRAS-tutorial-original`
1. created a new branch `CeRAS-tutorial` from the current master branch, and cherry-picked all the commits from `CeRAS-tutorial-original`
1. added a commit to take into account the changes of `operational-mission` that you did not have in your branch
1. checked results are identical between `CeRAS-tutorial` and `CeRAS-tutorial-original`. Hurrah!

At this point, your local branch `CeRAS-tutorial` should now match `original/CeRAS-tutorial-original`. In this case, you may safely checkout the new `original/CeRAS-tutorial` and overwrite your local branch.

Now for the review. GitHub did not allow me to annotate the notebook ("diff too large"), so I decided to create a `CeRAS-tutorial-review` branch where each commit is a proposal of change.
I suggest we discuss later about all proposed changes before merging.

Issue #265, opened 2021-02-04
Can we upgrade atmosphere function with better get_equivalent_airspeed/get_true_airspeed functions?


areysset:
Proposal:
[atmosphere.txt](https://github.com/fast-aircraft-design/FAST-OAD/files/5926899/atmosphere.txt)

Needed for CS23: we are currently patching the virtual environment!

christophe-david:
Please propose a PR with proper unit tests and write a clear rationale with it.

PR #281, opened 2021-03-22
Rework atmosphere

christophe-david:
#265 at least had the merit to make me realize that architecture of Atmosphere class missed consistency about speed parameters and could not be smoothly updated with a CAS computation:
- 2 functions to convert TAS from/to EAS
- a function to compute `unitary_reynolds`, that is actually equivalent to a speed once altitude is fixed, from Mach number
- no output for Mach number (class user had to use TAS and `speed_of_sound`. No big deal, ok, but still...)
In such architecture, was CAS to be computed from TAS? EAS ? Mach ?

This PR modifies the Atmosphere class to have four speed-related properties:
- `mach`
- `true_airspeed`
- `equivalent_airspeed`
- `unitary_reynolds`
Setting one of these properties will set values of the other ones (to avoid unnecessary computations, they are calculated on-demand, and then their value is stored.)
And then, adding `calibrated_airspeed` should be simple.

This PR proposes also a new location for the Atmosphere and FlightPoint classes, in fastoad.model_base, which makes more sense to me as these two are used by models (and actually, I even dreamed of embedding the atmosphere computation in the FlightPoint class, but I realized it is rather complex to do so...)


Martin-Delavenne (approved):
Good re-work of the atmosphere module.

PR #283, opened 2021-03-24
Mission fixes

christophe-david:
This PR fixes some details about the mission module:

- default time step for taxi segments was far too small
- duration of Breguet cruise segment was incorrectly computed.
- mission start is now better defined, as explained in the [updated documentation](https://fast-oad.readthedocs.io/en/mission-fixes/documentation/mission_module/mission_file/mission_file.html#mission-section)
- TOW was incorrectly computed using block fuel instead of fuel quantity at takeoff.


src/fastoad/models/performances/mission/openmdao/mission.py
```diff
@@ -408,6 +475,27 @@ def as_scalar(value):
...
             self.flight_points.to_csv(self.options["out_file"])
 
+    def _compute_taxi_out(self, inputs, outputs, propulsion_model):
+        """
+        Computes the taxi-out segment.
+        """
+        start_of_taxi_out = FlightPoint(
+            altitude=inputs[self._mission_vars.TAKEOFF_ALTITUDE.value],
+            true_airspeed=0.0,
+            # start mass is irrelevant here as long it does not get negative during computation.
```
Martin-Delavenne, line 485: Can't we compute the ramp weight on the same model as TOW computation through '_get_tow_component" to have a relevant weight here even if it would not change the results ? The ramp weight could then be added to the variable list which could be an interesting information.

Martin-Delavenne (approved):
Nice evolution with take-off and taxi out taking into account.  I added some suggestions that may not prevent from merging but could be improvement for futur developments.

PR #286, opened 2021-03-28
Trajectory post processing

ScottDelbecq:
This PR aims at providing a user interface within Jupyter notebooks for plotting and visualizing the different variables involved in the mission (performance module).
An example for the usage was added in the post-processing notebook:

trajectory_postprocessing

Units should in the mean time handled by the performance module. Hence, the mission viewer does not show variables' units.


christophe-david (approved):
Thanks!

PR #287, opened 2021-03-28
Issue 220 option documentation

### LINKED ISSUE #220: Documentation: describe the way OpenMDAO options are controlled from configuration file
The generated configuration file shows some examples, but this aspect should be addressed specifically in documentation

christophe-david:
This PR resolves #220.

It updates [documentation](https://fast-oad.readthedocs.io/en/issue-220_option-documentation/documentation/usage.html#problem-definition) about usage of OpenMDAO options, and enhances the way FAST-OAD provides help about available options.

BTW, registration of FAST-OAD models has been entirely put as decorators (`register.py` is deleted). This is done in dedicated commits, so this can be easily undone if we finally decide to keep `register.py`.


ScottDelbecq (approved):
Thanks for this precious documentation work.
All the changes in the vocabulary are very relevant.
I am ok with dropping the `register.py` file.

My only remark is that the .yml configuration file is outdated.
Is this PR the oportunity to update it and especially show an example with a subgroup?

christophe-david:
> Thanks for this precious documentation work.
> All the changes in the vocabulary are very relevant.
> I am ok with dropping the `register.py` file.
> 
> My only remark is that the .yml configuration file example given is outdated.
> Is this PR the oportunity to update it and especially show an example with a subgroup?

The PR is approved, so no more modification :)
Let's update the doc with the new fastoad.yml in a next small PR.

PR #288, opened 2021-03-30
Issue 245 jupyterlab

### LINKED ISSUE #245: variable_viewer toolbar not working in JupyterLab
The filtering toolbar of the variable_viewer is not working in Jupyter Lab but is in Jupyter Notebook

ScottDelbecq:
This PR aims at fixing the bug where the VariableViewer toolbar was not updating correctly.
Fixes #245


christophe-david (changes requested):
Nice! 
(For the records, that makes me realize we should probably have somewhere in documentation what should be installed for Jupyter lab. Having the information in the wiki is not enough for users)

Unfortunately, your branch is based on an old version of master branch (why?!), hence the merge conflict on dependency files, highlighted by GitHub.
I would suggest rebasing your branch on current master HEAD and doing again `poetry update`.
BTW, may I ask that you set `lxml` dependency to `^4.6.3` ? It would fix the last dependabot alert.

christophe-david (approved):

PR #289, opened 2021-03-30
Issue 280 n2 rendering

### LINKED ISSUE #280: 404 error when doing api.write_n2() in Jupyter notebook
OpenMDAO 3.7 introduces nice rendering of OpenMDAO views in Jupyter notebooks, but it collides with some features of FAST-OAD.
In particular, api.write_n2() shows a big frame with the "404: not found" message.

ScottDelbecq:
This PR is a work around for using `api.write_n2()`  introduced by openmdao 3.7 `om.n2()` display mecanism when using file absolute paths in Jupyter notebooks.

Later, we should probably use directly openmdao's `om.n2()` for displaying the n2 in Jupyter notebooks.

Fixes #280.

### DIFF
--- a/src/fastoad/cmd/tests/test_api.py
+++ b/src/fastoad/cmd/tests/test_api.py
@@ -101,10 +101,6 @@ def test_list_variables(cleanup):
 
 
 def test_write_n2(cleanup):
-    default_n2_file_path = "n2.html"
-    api.write_n2(CONFIGURATION_FILE_PATH)
-    assert pth.exists(default_n2_file_path)
-    os.remove(default_n2_file_path)
 
     n2_file_path = pth.join(RESULTS_FOLDER_PATH, "other_n2.html")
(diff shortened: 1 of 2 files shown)


src/fastoad/cmd/tests/test_api.py
```diff
@@ -101,7 +101,7 @@ def test_list_variables(cleanup):
 
 
 def test_write_n2(cleanup):
-    default_n2_file_path = pth.join(DATA_FOLDER_PATH, "n2.html")
+    default_n2_file_path = "n2.html"
```
christophe-david, line 104: The test for the default location of the generated file was useful because there was some path manipulation. But now it is a mere default value, I think it is not worth testing.
I would simply remove lines 104 to 107.

christophe-david (approved):
Thanks for the fix !

I'm Ok with the change. I just think some test code is no more relevant and could be removed, as stated in the inline comment.

PR #291, opened 2021-03-30
Units are now managed in FlightPoint class...

christophe-david:
... and inserted in column names of mission CSV files


ScottDelbecq (approved):
Top! Thanks!

PR #297, opened 2021-03-31
CeRAS harmonization

Martin-Delavenne:
This PR harmonizes reference data file for CeRAS reference aircraft within used within notebooks. 

All the entry were harmonized with respect to CeRAS_reference_date.xml file values except TLAR range and Payload and mission values. Indeed, in the reference file for CeRAS case study 3 different missions are described that are not necessary for tutorial notebooks to run properly.


Martin-Delavenne:
I updated the integration test with correct values as tests failed during the first push. I noticed that the timestep configuration file did not incorporate the wing position module, therefore for consitency with the Breguet computation I modified the timstep configuration file accordingly.

christophe-david (approved):
Good to have now this consistency !

I also added a commit to put a note for explaining the warning message about T4 value. We should have added it long ago, so I decided to ass it here.

Issue #306, opened 2021-04-12
Upgrade to FAST-OAD 1.0.0 not successfull


AleSgueglia:
Hello, 
I open this issue because I tried to update my fast-oad version to the latest release. 
Through the Anaconda prompt I run "pip install --upgrade fast-oad" but the code is not installed. 
The following error appears:  

"Attempting uninstall: ruamel.yaml
    Found existing installation: ruamel-yaml 0.15.87
ERROR: Cannot uninstall 'ruamel-yaml'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall."

I tried to manually uninstall the package ruamel.yaml but this does not solve the issue. 

Thanks in advance for any help to solve the issue. 

Regards,
Alessandro Sgueglia

christophe-david:
Hi,

It looks like Anaconda comes with a "clone" of `ruamel.yaml` (`ruamel_yaml` and a kind of alias `ruamel-yaml`). Apparently, it is not a problem with the old version of Anaconda I had, but with the latest one, I get the same problem as you. It looks impossible to either upgrade or uninstall this package and IMHO, it is a bug in Anaconda install.

As a workaround, I would advise that you create a new virtual environment (possible through command line, but you may also do it through GUI using Anaconda navigator, see image below).
Actually, I think this a good solution, even without the `ruamel.yaml` problem, especially because FAST-OAD brings some big [changes](https://github.com/fast-aircraft-design/FAST-OAD/releases/tag/v1.0.0) compared to 0.5.4b. 
Then the advantage of the virtual environment is double:
- you keep your current working environment (with FAST-OAD 0.5.4b) untouched
- you will not get this `pip` error, since `ruamel-yaml` will not be in your new environment.

_____
_____
_Virtual environment though anaconda navigator..._

<img width="1016" alt="conda" src="https://user-images.githubusercontent.com/49477099/114355152-6a2de580-9b6f-11eb-96f4-3e2534d13e2f.png">

younes-jc:
The solution @christophe-david mentioned worked for me. However, when trying to run any of the notebooks I get the following error (for the image below, I tried to execute the first code cell of "01_tutorial", but the same error appears for "CeRAS_case_study" notebook):

![image](https://user-images.githubusercontent.com/82504418/114856458-6c9a7480-9de7-11eb-9a70-bc1e28ab3e1b.png)

[...]

PR #308, opened 2021-04-13
Dependency update

christophe-david:
This PR brings several changes related to project dependencies:

-  Dependency to `jupyterlab` is removed. In documentation, we advertise `jupyter-notebook` and we never mention `jupyterlab`. We also know that usage of jupyterlab needs additional operations from user for our GUI elements to work. So, there is no reason to force its installation in user environment. It does not prevent a user to have it installed, and it does not prevent us to ensure compatibility with it (BTW, we indeed have widget problems with Jupyter 3.0+)
- Dependency to `ruamel.yaml` now requires a lower version (the lowest one I could install on my machine). It should solve #306 by not triggering an update of `ruamel.yaml` in Anaconda (but still, the bug is Anaconda's 😠)
- Our "watchman test" workflow has been changed: I observed that `poetry update` did not bring the latest version of `pandas`. The reason is that latest `pandas` requires 3.7.1, though we only require 3.7+ in our `pyproject.toml`. So Poetry keeps `pandas` compatible with 3.7.0. Therefore, to have the watchman tests use the very latest versions of all packages, I realized the best way was to not use Poetry, but a mere `pip install .`


.github/workflows/watchman_tests.yml
```diff
@@ -23,27 +23,16 @@ jobs:
...
-      - name: Update production dependencies
-        run: poetry update --no-dev
-        shell: bash
-
-      - name: Install dependencies
-        run: poetry install
+      - name: Install # installing pytest-cov is needed because of pytest.ini
+        run: |
+          pip install .
+          pip install pytest pytest-cov nbval
```
ScottDelbecq, line 29: How does pip work without setup.py nor requirements.txt?

ScottDelbecq (approved):

.github/workflows/watchman_tests.yml
christophe-david: pip uses the `pyproject.toml` file (as stated [here](https://pip.pypa.io/en/stable/reference/pip/#build-system-interface)). `setup.py` is required for installing in editable mode (`-e` option), but for that, we just do `poetry install`.

Issue #328, opened 2021-05-05
input file not overwritten when a file with the same name already exists 


Martin-Delavenne:
if an input file (with the same name as the one in the conf file) is already present in the working directory calling `oad.generate_inputs(input_file_path, overwritte = True)` does not erase the existing file and replace it by the new but only add variables which are different. 

This may lead to awkward situations for which the existing input file may contain variables not related to the models currently under study then calling `oad.generate_inputs` while result in an input file with unecessary variables. It could confuse the user. 

For example for the very simple "model" below: 
```python 
@oad.RegisterOpenMDAOSystem("fastoad.my_comp")
class Comp1(om.ExplicitComponent):
    def setup(self):
        self.add_input("x", desc="first variable", units="m")
        self.add_input("y", desc="second variable", units="m")
        
        self.add_output("z", desc="output variable", units="m**2")
    
    def compute(self, inputs, outputs):
        x = inputs["x"]
        y = inputs["y"]

        outputs["z"] = x * y
```

if no input file exists the result of `oad.generate_inputs(input_file_path, overwritte = True)` is 
```xml
<FASTOAD_model>
  <x units="m" is_input="True">1.0</x>
  <y units="m" is_input="True">1.0</y>
</FASTOAD_model>
```
Which is the expected behaviour. 
But if an input file with the same name as the one indicated in the configuration file exists but with totally different variables : 
```xml
<FASTOAD_model>
  <var_from_other_model_1 units="m" is_input="True">1.0</var_from_other_model_1>
  <var_from_other_model_2 units="m" is_input="True">1.0</var_from_other_model_2>
</FASTOAD_model>
```

The results of `oad.generate_inputs(input_file_path, overwritte = True)` is: 
```xml
<FASTOAD_model>
  <var_from_other_model_1 units="m" is_input="True">1.0</var_from_other_model_1>
  <var_from_other_model_2 units="m" is_input="True">1.0</var_from_other_model_2>
  <x units="m" is_input="True">1.0</x>
  <y units="m" is_input="True">1.0</y>
</FASTOAD_model>
```

christophe-david:
Problem confirmed and reproduced.

PR #330, opened 2021-05-07
Fix overwrite of input file

christophe-david:
This PR solves #328.

The `DataFile` class, now used in `generate_inputs` was intended to load the content of a data file if it existed.
The class constructor has been modified to allow the data not to be loaded if necessary.


Martin-Delavenne (approved):
Thank you for this PR. The correction makes the behaviour sounder when an input file is already existing in the working directory.

PR #333, opened 2021-05-10
Allowing "useless" sections in configuration file

### LINKED ISSUE #324: Syntax verification is rigid
Hello,
We want to add sections into the YAML configuration file so that we can use the same file for different purposes but the same problem. 

The fact is that the current syntax verification do not allow to add sections other than [model] or [optimization], for example.

As long as the sections added are not used by FAST-OAD itself (and not even FAST-OAD sections), it could be nice that the verification gives a _Warning_ and not an _Error_. 

Thank you

christophe-david:
This PR closes #324.

Unknown keys are now accepted in the configuration file (at root level).
To keep being informative about possible typos, a warning is issued for each found unknown key.


ScottDelbecq (approved):
Thank you for this PR.
When testing I was wondering if it is not worth to have a function in our api that loads the configuration file?

PR #335, opened 2021-05-16
Sub-model registering 1

christophe-david:
This PR competes with PR #336 to implements the idea in #255, that is to provide a way of replacing only a part of a FAST-OAD module.

Typically, a FAST-OAD module is an OpenMDAO group that uses several components as subsystems.
Instead of doing:
```python
class MyGroup(om.Group):
    def setup():
        self.add_subsystem(
            "system1", MyClassForSystem1(), promotes=["*"],
        )
```
We can now do:
```python
class MyGroup(om.Group):
    def setup():
        self.add_subsystem(
            "system1", RegisterSubmodel.get_submodel("service.for.system.1"), promotes=["*"],
        )
```

In above code `RegisterSubmodel.get_submodel("service.for.system.1")` asks for an instance of a class that has been registered as providing the service `"service.for.system.1"`.

The registering is done with the `RegisterSubmodel` decorator:
```python
@RegisterSubmodel("service.for.system.1", "id.of.this.component.that.provides.the.system.1.service")
class MyClassForSystem1(om.Group):
    ...
```

As you can see, what is asked for in the `add_subsystem` above is the service, not a specific component.
The choice of the component that will be provided is identical to #336, and depends on conditions:
- if only one component is registered for providing the service, the component is used
- if several components are registered for the same service, the decision come from the dictionary `RegisterSubmodel.active_models` that can associate a service (dict keys) to the component to use (dict items). If the service is not in the dict, an error is raised. `RegisterSubmodel.active_models` is empty by default and can be fed two ways:
   - directly in Python code
   - In the configuration file, with the section `submodels:`. If `RegisterSubmodel.active_models` is modified in the `initialize()` method of any FAST-OAD module, it will be overridden by the configuration file.

Compared to #336, the current proposal adds the `RegisterSubmodel` class, that is independent of the existing  `RegisterService` class. 
The registering process do not require to define specialized decorators beforehand.
The definition of the service identifier is done at registering time. This allows to use only one decorator, `RegisterSubmodel`, but forces to repeat the service identifier each time.
Also, this solution does not provide a way to specify an interface for the registered submodels. If it was needed, it could be done only when "asking" for the submodels (in the example above, in the `add_subsystem()`) and then, checking that the interface is respected could be done only at runtime.

A real case example is available in draft PR #337.


Martin-Delavenne:
Thank you for the proposal of this two methods to register submodels. My preference goes to this one that I found more directly accessible for users using just one decorator. I find also the formalism more compact and easy to read than in #336 where decorator class must be define beforehand.

ScottDelbecq:
I agree with @Martin-Delavenne that this is the more user friendly manner for registering sub-models. I would go for that as well. However, I wonder how do we choose the sub-model we want from the configuration file? Like **Library of OpenMDAO systems** in #255?

PR #336, opened 2021-05-17
Sub-model registering 2

christophe-david:
This PR competes with PR #335 to implements the idea in #255, that is to provide a way of replacing only a part of a FAST-OAD module.

Typically, a FAST-OAD module is an OpenMDAO group that uses several components as subsystems.
Instead of doing:
```python
class MyGroup(om.Group):
    def setup():
        self.add_subsystem(
            "system1", MyClassForSystem1(), promotes=["*"],
        )
```
We can now do:
```python
class RegisterSystem1Service(RegisterSubmodel, service_id="service.for.system.1"):
    """Register models for computing mass of wings."""

class MyGroup(om.Group):
    def setup():
        self.add_subsystem(
            "system1", RegisterSystem1Service.get_submodel(), promotes=["*"],
        )
        ...
```

In above code `RegisterSystem1Service.get_submodel()` asks for an instance of a class that has been registered as providing the service `"service.for.system.1"`.

The registering is done with the defined decorator:
```python
@RegisterSystem1Service("id.of.this.component.that.provides.the.system.1.service")
class MyClassForSystem1(om.Group):
    ...
```
As you can see, what is asked for in the `add_subsystem` above is the service, not a specific component.
The choice of the component that will be provided is identical to #335, and depends on conditions:
- if only one component is registered for providing the service, the component is used
- if several components are registered for the same service, the decision come from the dictionary `RegisterSubmodel.active_models` that can associate a service (dict keys) to the component to use (dict items). If the service is not in the dict, an error is raised. `RegisterSubmodel.active_models` is empty by default and can be fed two ways:
   - directly in Python code
   - In the configuration file, with the section `submodels:`. If `RegisterSubmodel.active_models` is modified in the `initialize()` method of any FAST-OAD module, it will be overridden by the configuration file.

Compared to #335, the current proposal relies on the already existing `RegisterService` class (`RegisterSubmodel` simply inherits from it). The definition of the decorators is a bit more verbose, but a bit less prone to typos, as the service identifier is used only once in Python, when defining the decorator class. Also, this proposal already offers a way to specify a base class that should be parent to decorated class, which allows to impose a given interface, which could be useful to ensure correct interaction between sub-models.

A real case example is available in draft PR #338.


christophe-david:
Method 1 has been preferred by the team.
Closing this PR.

PR #332, opened 2021-05-10
Issue 313 input generation

christophe-david:
This PR solves #313

PR #341, opened 2021-05-20
Issue 334 - local inputs connected to promoted variables

### LINKED ISSUE #334: Connected promoted inputs are considered as global inputs
There is an problem when determining wether if a variable is a global input of the problem when using the connect() mechanism with the promotion mechanism. In the example below, `y1` is an output of `dis1` and an input of `functions` which are both connected. Hence, `y1.is_input` should be `False`. 

```python
import numpy as np
import openmdao.api as om
from fastoad.openmdao.variables import VariableList


class Disc1(om.ExplicitComponent):
    """ An OpenMDAO component to encapsulate Disc1 discipline """

    def setup(self):
        self.add_input("x", val=np.nan, desc="")  # NaN as default for testing connexion check
        self.add_input("z", val=[5, 2], desc="", units="m**2")  # for testing non-None units
        self.add_input("y2", val=1.0, desc="")

        self.add_output("y1", val=1.0, desc="")
        self.declare_partials("*", "*", method="fd")

    def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
        """
        Evaluates the equation
        y1 = z1**2 + z2 + x1 - 0.2*y2
        """
        z1 = inputs["z"][0]
        z2 = inputs["z"][1]
        x1 = inputs["x"]
        y2 = inputs["y2"]

        outputs["y1"] = z1 ** 2 + z2 + x1 - 0.2 * y2
        
        
class Disc2(om.ExplicitComponent):
    """ An OpenMDAO component to encapsulate Disc2 discipline """

    def setup(self):
        self.add_input("z", val=[5, 2], desc="", units="m**2")  # for testing non-None units
        self.add_input("y1", val=1.0, desc="")

        self.add_output("y2", val=1.0, desc="")
        self.declare_partials("*", "*", method="fd")

    def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
        """
        Evaluates the equation
        y2 = y1**(.5) + z1 + z2
        """

        z1 = inputs["z"][0]
        z2 = inputs["z"][1]
        y1 = inputs["y1"]

        # Note: this may cause some issues. However, y1 is constrained to be
        # above 3.16, so lets just let it converge, and the optimizer will
        # throw it out
        if y1.real < 0.0:
            y1 *= -1

        outputs["y2"] = y1 ** 0.5 + z1 + z2


class Functions(om.ExplicitComponent):
    """ An OpenMDAO component to encapsulate Functions discipline """

    def setup(self):
        self.add_input("x", val=2, desc="")
        self.add_input(
            "z", val=[np.nan, np.nan], desc="", units="m**2"
        )  # NaN as default for testing connexion check
        self.add_input("y1", val=1.0, desc="")
        self.add_input("y2", val=1.0, desc="")

        self.add_output("f", val=1.0, desc="")

        self.add_output("g1", val=1.0, desc="")

        self.add_output("g2", val=1.0, desc="")
        self.declare_partials("*", "*", method="fd")

    def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
        """ Functions computation """

        z2 = inputs["z"][1]
        x1 = inputs["x"]
        y1 = inputs["y1"]
        y2 = inputs["y2"]

        outputs["f"] = x1 ** 2 + z2 + y1 + exp(-y2)
        outputs["g1"] = 3.16 - y1
        outputs["g2"] = y2 - 24.0

group = om.Group()
group.add_subsystem("disc1", Disc1(),  promotes=["x", "z"])
group.add_subsystem("disc2", Disc2(),  promotes=["z"])
group.add_subsystem("functions", Functions(),  promotes=["*"])

group.connect("disc1.y1", "disc2.y1")
group.connect("disc2.y2", "disc1.y2")
group.connect("disc1.y1", "y1")
group.connect("disc2.y2", "y2")

problem = om.Problem(group)
problem.setup()

vars = VariableList.from_problem(problem, use_initial_values=False, get_promoted_names=True)

# Should be False
print(vars["y1"].is_input)
```

ScottDelbecq:
This PR fixes #334 (partly solved by #332) where some variables are tagged as inputs instead of outputs when using `VariableList.from_problem()`.

To fix this, `VariableList.from_problem()` was modified to handle the effect of using `connect()` with local inputs and promoted outputs. This behaviour can be observed when using `list_variables()` or the `variable_viewer()`.

However, `generate_inputs()` works fine and gives the correct input list because it uses `_utils.get_unconnected_input_names()`. This brings up the fact that `VariableList.from_problem()` is too complicated for what we really need it to do and should deserve a rework in a later PR.

### DIFF
--- a/src/fastoad/openmdao/variables.py
+++ b/src/fastoad/openmdao/variables.py
@@ -576,6 +576,15 @@ class VariableList(list):
                 if "." not in metadata["prom_name"]
             }
 
+            if get_promoted_names:
+                # Check connections
+                for name, metadata in inputs.copy().items():
+                    source_name = problem.model.get_source(name)
+                    # If an local input is connected to a promoted output
+                    if not source_name.startswith("_auto_ivc.") and source_name != name:
+                        del inputs[name]
+                        outputs[name] = metadata
+
         # Add "is_input" field
         for metadata in inputs.values():
             metadata["is_input"] = True
@@ -587,6 +596,13 @@ class VariableList(list):
             final_inputs = inputs
             final_outputs = outputs
         else:
+            # Remove from inputs the variables that are outputs of some other component
+            promoted_inputs = {
+                metadata["prom_name"]: dict(metadata, is_input=True)
+                for metadata in inputs.values()
+                # if metadata["prom_name"] not in promoted_outputs
+            }
+
             promoted_outputs = {}
             for metadata in outputs.values():
                 prom_name = metadata["prom_name"]
@@ -618,15 +634,15 @@ class VariableList(list):
                         # We already have a non-NaN value and current variable provides no unit.
                         # No need for using the current variable.
                         continue
+                if prom_name not in promoted_inputs:
+                    promoted_outputs[prom_name] = metadata
 
-                promoted_outputs[prom_name] = metadata
-
-            # Remove from inputs the variables that are outputs of some other component
-            promoted_inputs = {
-                metadata["prom_name"]: dict(metadata, is_input=True)
-                for metadata in inputs.values()
-                if metadata["prom_name"] not in promoted_outputs
-            }
+            # # Remove from inputs the variables that are outputs of some other component
+            # promoted_inputs = {
+            #     metadata["prom_name"]: dict(metadata, is_input=True)
+            #     for metadata in inputs.values()
+            #     if metadata["prom_name"] not in promoted_outputs
+            # }
 
             final_inputs = promoted_inputs
             final_outputs = promoted_outputs
--- a/src/fastoad/openmdao/variables.py
+++ b/src/fastoad/openmdao/variables.py
@@ -638,13 +638,6 @@ class VariableList(list):
                 if prom_name not in promoted_inputs:
                     promoted_outputs[prom_name] = metadata
 
-            # # Remove from inputs the variables that are outputs of some other component
-            # promoted_inputs = {
-            #     metadata["prom_name"]: dict(metadata, is_input=True)
-            #     for metadata in inputs.values()
-            #     if metadata["prom_name"] not in promoted_outputs
-            # }
-
             final_inputs = promoted_inputs
             final_outputs = promoted_outputs
 
--- a/src/fastoad/openmdao/variables.py
+++ b/src/fastoad/openmdao/variables.py
@@ -580,8 +580,9 @@ class VariableList(list):
                 # Check connections
                 for name, metadata in inputs.copy().items():
                     source_name = problem.model.get_source(name)
-                    # If an local input is connected to a promoted output
                     if not source_name.startswith("_auto_ivc.") and source_name != name:
+                        # This variable is connected to another variable of the problem: it is
+                        # not an actual problem input. Let's move it to outputs.
                         del inputs[name]
                         outputs[name] = metadata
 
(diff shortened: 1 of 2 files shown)


src/fastoad/openmdao/variables.py
```diff
@@ -576,6 +576,15 @@ def from_problem(
                 if "." not in metadata["prom_name"]
             }
 
+            if get_promoted_names:
+                # Check connections
+                for name, metadata in inputs.copy().items():
```
christophe-david, line 581: Simple comment, no change needed:
Generally speaking, I try to avoid copying items as much as I can.
I understand why you iterate on `inputs.copy()` instead of `inputs`, but I don't think it is good practice, because it makes you iterate twice on the dict (once for the copy, and then for your loop).
A better solution would probably be to iterate on the original dict and build a list of the variables that should be transferred from inputs to outputs, and then do the transfer by iterating this list (that will be far smaller that the dict).
Anyway, there is no performance issue in this part of the code, so we can live with that.

christophe-david (changes requested):
That works Ok, thanks.

A remark on the order of commits: it is better to implement the test, then the fix. It makes easier for the reviewer to checkout the commit with the test, see that it fails, then checkout the head of the branch and see that the test passes.
Even if you did not work in this order (though it is generally advised to do so), an interactive rebase allows to put your commit history in the right order.

I just ask for a change in a comment and it will be Ok.

src/fastoad/openmdao/variables.py
```diff
@@ -576,6 +576,15 @@ def from_problem(
...
+                    source_name = problem.model.get_source(name)
+                    # If an local input is connected to a promoted output
+                    if not source_name.startswith("_auto_ivc.") and source_name != name:
+                        del inputs[name]
```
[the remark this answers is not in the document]
ScottDelbecq: I agree. Thanks! Corrected in fcd87be.

src/fastoad/openmdao/variables.py
ScottDelbecq: Noted, I had thought about that...

ScottDelbecq:
Thanks for you comments. Can we integrate this PR in #340?

src/fastoad/openmdao/variables.py
```diff
@@ -618,15 +635,15 @@ def from_problem(
...
-            }
+                if prom_name not in promoted_inputs:
+                    promoted_outputs[prom_name] = metadata
+
+            # # Remove from inputs the variables that are outputs of some other component
+            # promoted_inputs = {
+            #     metadata["prom_name"]: dict(metadata, is_input=True)
+            #     for metadata in inputs.values()
+            #     if metadata["prom_name"] not in promoted_outputs
+            # }
```
christophe-david, line 646: I missed that in my previous reading. Could you please remove these lines ?

christophe-david (changes requested):
Thanks for the changes.
Sorry, in the previous review, I missed you commented out some code (that, indeed, should legitimately be removed).

src/fastoad/openmdao/variables.py
ScottDelbecq: Thanks. Removed in fbe8bbb.

christophe-david (approved):

christophe-david:
The change has been integrated in #340.
Looking forward your review :)

PR #345, opened 2021-06-03
Extensible enum classes

### LINKED ISSUE #342: Add alternative engine settings
Hello, 

I am looking for an easy way to add some engine settings to the ones already defined. I tried inheriting the class, but it doesn't let me add another engine setting. I get an error which says : "Cannot extend enumerations". Is there a specific method which does that?

Thanks,
Vincenzo

christophe-david:
This PR resolves #342 by using the [`aenum`](https://github.com/ethanfurman/aenum) library instead of standard [`enum`](https://docs.python.org/3/library/enum.html).

Now following enum classes are extensible:
- EngineSetting
- FlightPhase
- RangeCategory
- PolarType
- ModelDomain

____

`aenum` has also been used so that FlightSegment-derived classes auto-declare the name that will identify them in mission definition files. This should make easy the development of new flight segments.

For example, `AltitudeChangeSegment` is now declared with:
```python
class AltitudeChangeSegment(ManualThrustSegment, mission_file_keyword="altitude_change"):
```
And that is enough to have it available in mission definition.

____

Finally, it has been the occasion to realize that the usage of `Enum` in `MissionComponent` class was a bit awkward, and that `namedtuple` was probably a better choice.


ScottDelbecq (approved):
Thanks for this PR which will be very useful!

PR #351, opened 2021-06-10
Added Binder image

christophe-david:
This PR allows to use FAST-OAD notebooks on a [Binder](https://mybinder.org) page.

- The main change is that a `binder/` folder has been added with needed configuration files, as instructed [here](https://mybinder.readthedocs.io/en/latest/using/config_files.html).
   So, after the merge in `master` branch, it will be possible to have a Binder session for any branch or tag with following lines (modify _$branch-name$_ by the branch or tag name):
   - for Jupyter Notebook: https://mybinder.org/v2/gh/fast-aircraft-design/FAST-OAD.git/$branch-name$/filepath=src%2Ffastoad%2Fnotebooks
   - for Jupyter Lab: https://mybinder.org/v2/gh/fast-aircraft-design/FAST-OAD.git/$branch-name$?urlpath=lab%2Ftree%2Fsrc%2Ffastoad%2Fnotebooks
 
   Therefore, an image of this branch can be played with at https://mybinder.org/v2/gh/fast-aircraft-design/FAST-OAD.git/test-binder?urlpath=lab%2Ftree%2Fsrc%2Ffastoad%2Fnotebooks

- Binder will build the image of the branch/tag on-demand. The build process is rather long (a couple of minutes). But if an image has already been built, it will be reused (in a new session anyway, but the build process won't happen. Therefore, it looked good to pre-build some images. So some additions have be made to GitHub workflows to ensure that a Binder image is built for:
    - each commit in master branch
    - each update of `latest-release` tag, knowing that the build workflow ensures that this tag is moved to match each new release tag.

- `README.md` has been modified to provide a link for a Jupyter Lab session on `latest-release` tag

> **Note:**
> **In the future, we may want to use our own BinderHub server, to ensure that our images are always available, and maybe quicker to launch.**
> **The documentation for that is here: https://binderhub.readthedocs.io/en/latest/**

____

2 non-related changes are also in this PR:
- `pyproject.toml` has been modified to have JupyterLab as optional dependency, so we may install it in our projects with:
    ```
    poetry install -E jupyterlab
    ```
    It allows to specify the version we want for JupyterLab (i.e. not 3.+) without making user installing it needlessly. It should also allow users to install the needed version of JupyterLab with:
   ```
    pip install FAST-OAD[jupyterlab]
    ```
   _Actually, I made this change as I was hoping to produce requirements.txt for Binder, but automating that happened to be a bit complicated (because this `requirements.txt` must contain FAST-OAD). And the simple `requirements.txt` I have put looks good enough._

- Additionally, the test workflows do not trigger anymore on pull requests. The idea was to ensure that notebooks are always tested before merge, but Actions are made so that the workflow will happen on pull requests AND push, though we would rather have one replacing the other. That triggered too many workflows.
  The current version of the workflow still allows to force manually that notebooks are tested, and also tests notebooks on each commit on `master` branch.


binder/postBuild
```diff
@@ -0,0 +1,4 @@
+jupyter labextension install @jupyter-widgets/jupyterlab-manager --no-build
+jupyter labextension install ipysheet --no-build
+jupyter labextension install jupyterlab-plotly --no-build
+jupyter lab build --dev-build=False --minimize=False
```
ScottDelbecq, line 4: I just realized that to make the `MissionViewer.display()` work in Jupyter Lab you have to install `jupyter labextension install plotlywidget` since it uses a plotly `FigureWidget`. Ashame it is not covered by `jupyterlab-plotly`.

ScottDelbecq (changes requested):
Thanks for this very useful PR!
I greatly approve to use Jupyter Lab as the default environment for Binder link in the README.
However, to make the `MissionViewer` work properly an other jupyterlab extension has to be installed (see my single comment). Otherwise everything seems to run very smoothly!

Also, once [ipysheet will be compatible with Jupyter Lab 3](https://github.com/QuantStack/ipysheet/issues/184) it will be worth to test and move our jupyterlab dependency to at least version 3.

binder/postBuild
christophe-david: Thanks for the advice.
Done in cad824984a6492882318d4d948f45a7e723b4e93.

ScottDelbecq (approved):
Just tested it and it's working perfectly :+1: 
Thanks!

Martin-Delavenne (approved):
Thanks for this PR. 

I tested, works fine.

PR #368, opened 2021-07-22
Issue 311 - Incorrect taxi-in distance

### LINKED ISSUE #311: Incorrect distance computation for taxi-in phase
In FAST-OAD 1.0, as pointed out in #309, taxi-in phase is computed with the last achieved speed, that is normally a flight speed, hence far too fast.

It results in a far too large covered distance, and probably affects the fuel consumption for this phase.

ScottDelbecq:
This PR fixes #311.
It corrects a far too large assessment of the distance during the taxi-in segment. This was due to the fact that the last FlightPoint used was from the holding phase so a quite high mach applied to the duration of the taxi-in segment.

Now taxi segments have a distance which is equal to `0` since we now set the `true_airspeed` parameter of this segment to `0`. Changing the `true_airspeed` has lead to a lower `sfc` hence different fuel mass in the taxi-in segment. This change in mass has spread in the different loops which required an update of the variable values (MTOW, wing_area...) in the tests.

### DIFF
--- a/src/fastoad/models/performances/mission/mission_definition/tests/data/mission.yml
+++ b/src/fastoad/models/performances/mission/mission_definition/tests/data/mission.yml
@@ -121,12 +121,14 @@ phases:
     parts:
       - segment: taxi
         thrust_rate: ~
+        true_airspeed: 0.
         target:
           time: ~duration
   taxi_in:
     thrust_rate: ~
     parts:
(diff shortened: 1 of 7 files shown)


christophe-david (approved):
Thanks for this update (and for the small fix about thrust_rate unit !)

PR #369, opened 2021-07-26
Issue 365 listing custom modules

### LINKED ISSUE #365: Listing custom modules without a configuration file
When custom modules are developed, it would be interesting to check that they are correctly registered without having to write a configuration file.

In CLI and in Python API, the command for listing modules uses a configuration file as input. Alternatively, it could be fed with a list of folders.

**Additional context**
Of course, this feature request does not apply when custom modules are declared as plugins, because they will be visible when listing modules without providing any input to the command.

ScottDelbecq:
This PR closes #365. It provides the possiblity to use `list_modules()` without providing a configuration file in the presence of custom modules. 
Now (but this is open to discussion), the source used for `list_modules()` can be:
- the path of a configuration file
- the path of a folder containing the custom modules
- a list of folder paths that each contain custom modules

### DIFF
--- a/src/fastoad/cmd/api.py
+++ b/src/fastoad/cmd/api.py
@@ -221,20 +221,24 @@ def list_modules(
     if out is None:
         out = sys.stdout
 
-    if source_path:
-        if isinstance(source_path, str) and pth.isfile(source_path):
+    if isinstance(source_path, str):
+        if pth.isfile(source_path):
             conf = FASTOADProblemConfigurator(source_path)
             conf._set_configuration_modifier(_PROBLEM_CONFIGURATOR)
             # As the problem has been configured,
             # BundleLoader now knows additional registered systems
-        elif isinstance(source_path, str) and pth.isdir(source_path):
+        elif pth.isdir(source_path):
             RegisterOpenMDAOSystem.explore_folder(source_path)
-        elif any(pth.isdir(x) for x in source_path):
-            for folder_path in source_path:
-                if not pth.exists(folder_path):
-                    _LOGGER.warning("SKIPPED %s: it does not exist.", folder_path)
-                else:
-                    RegisterOpenMDAOSystem.explore_folder(folder_path)
+        else:
+            raise FileNotFoundError("Could not find %s" % source_path)
+    elif isinstance(source_path, list):
+        for folder_path in source_path:
+            if not pth.isdir(folder_path):
+                _LOGGER.warning("SKIPPED %s: folder does not exist.", folder_path)
+            else:
+                RegisterOpenMDAOSystem.explore_folder(folder_path)
+    elif source_path is not None:
+        raise RuntimeError("Unexpected type for source_path")
 
     if verbose:
         cell_list = _get_detailed_system_list()
--- a/src/fastoad/cmd/api.py
+++ b/src/fastoad/cmd/api.py
@@ -196,7 +196,7 @@ def list_variables(
 
 
 def list_modules(
-    configuration_file_path: str = None,
+    source_path: str = None,
     out: Union[IO, str] = None,
     overwrite: bool = False,
     verbose: bool = False,
@@ -204,10 +204,10 @@ def list_modules(
 ):
     """
     Writes list of available systems.
-    If configuration_file_path is given and if it defines paths where there are registered systems,
+    If source_path is given and if it defines paths where there are registered systems,
     they will be listed too.
 
-    :param configuration_file_path:
+    :param source_path: either a configuration file path, folder path, or list of folder path
     :param out: the output stream or a path for the output file (None means sys.stdout)
     :param overwrite: if True and out is a file path, the file will be written even if one already
                       exists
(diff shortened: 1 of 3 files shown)


src/fastoad/cmd/api.py
```diff
@@ -221,10 +221,20 @@ def list_modules(
...
+            # As the problem has been configured,
+            # BundleLoader now knows additional registered systems
+        elif isinstance(source_path, str) and pth.isdir(source_path):
+            RegisterOpenMDAOSystem.explore_folder(source_path)
+        elif any(pth.isdir(x) for x in source_path):
+            for folder_path in source_path:
+                if not pth.exists(folder_path):
+                    _LOGGER.warning("SKIPPED %s: it does not exist.", folder_path)
+                else:
+                    RegisterOpenMDAOSystem.explore_folder(folder_path)
```
christophe-david, line 237: In your code, there is a first check than the list contains a folder path, then a check on each item that it is an existent path. I see 3 problems here:
- there is an assumption that, at this point, `source_path` is a list
- `pth.isdir` checks path existence, so by checking it again in the `for` loop, the job is done twice
- as the second check is only for existence, if `source_path` contains an existing folder path AND an existing file path, `explore_folder` will be fed also with the file path... and that should be bad.

Therefore I suggest:
```suggestion
        elif isinstance(source_path, list):
            for folder_path in source_path:
                if not pth.isdir(folder_path):
                    _LOGGER.warning("SKIPPED %s: folder does not exist.", folder_path)
                else:
                    RegisterOpenMDAOSystem.explore_folder(folder_path)
        else:
            raise RuntimeError("Unexpected type for source_path")
```

Here again, I end with a `else:`, but I did not put much effort in the error handling, as I find this error unlikely.

src/fastoad/cmd/fast.py
```diff
@@ -275,6 +274,12 @@ def run(self):
             action="store_true",
             help="shows detailed information for each system",
         )
+        parser_list_modules.add_argument(
+            "source_path",
+            nargs="?",
```
christophe-david, line 279: For the command to accept several arguments, you have to use:
```suggestion
            nargs="*",
```
See https://docs.python.org/3/library/argparse.html#nargs

christophe-david (changes requested):
Thank you for this update.

I am OK with the principle, but I think some details need to be changed.

src/fastoad/cmd/api.py
ScottDelbecq: Thanks for the suggestions! This new `else` means we do not allow to not provide any source_path anymore `source_path=None`.
Do you think we should keep it like that?

christophe-david: You got a point. I forgot we need to keep the possibility to have `None `as `source_path`.
Changing `else:` by `elif source_path is not None:` should do the trick, shouldn't it ?

ScottDelbecq: I agree! I would also keep the `else` to avoid unexpected types of entries. Do you agree?

christophe-david: Actually, my suggestion stand for the `else:` because the only unprocessed case will be that source_path is None, and we want to do nothing in this case.

If you prefer to do it by the book (and I can agree with that), you may replace the `else:` by
```python
elif source_path is None:
    pass # nothing needs to be done
else:
```

This is equivalent to my previous suggestion but more expressive.

ScottDelbecq: I agree that is why in the mean time I deleted my comment :)

src/fastoad/cmd/api.py
```diff
@@ -196,18 +196,18 @@ def list_variables(
 
 
 def list_modules(
-    configuration_file_path: str = None,
+    source_path: str = None,
```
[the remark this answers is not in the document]
ScottDelbecq: Corrected in c2f7a96

src/fastoad/cmd/fast.py
ScottDelbecq: Corrected in ff70f99

src/fastoad/cmd/api.py
```diff
@@ -221,10 +221,20 @@ def list_modules(
...
-        conf._set_configuration_modifier(_PROBLEM_CONFIGURATOR)
-    # As the problem has been configured, BundleLoader now knows additional registered systems
+    if source_path:
+        if isinstance(source_path, str) and pth.isfile(source_path):
+            conf = FASTOADProblemConfigurator(source_path)
+            conf._set_configuration_modifier(_PROBLEM_CONFIGURATOR)
+            # As the problem has been configured,
+            # BundleLoader now knows additional registered systems
+        elif isinstance(source_path, str) and pth.isdir(source_path):
+            RegisterOpenMDAOSystem.explore_folder(source_path)
```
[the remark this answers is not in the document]
ScottDelbecq: Corrected in ff59930

src/fastoad/cmd/api.py
ScottDelbecq: Corrected in ff59930

christophe-david (approved):
Thank you for the update

PR #371, opened 2021-07-26
Atmosphere cas added

areysset:
The method calculating CAS in atmosphere class has been added.
The corresponding test has also been added.

### DIFF
--- a/src/fastoad/model_base/atmosphere.py
+++ b/src/fastoad/model_base/atmosphere.py
@@ -185,44 +185,37 @@ class Atmosphere:
                 self._true_airspeed = self._unitary_reynolds * self.kinematic_viscosity
             if self._calibrated_airspeed is not None:
                 cas = np.array(self._calibrated_airspeed)
-                self._true_airspeed = np.zeros(np.size(cas))
                 sea_level = Atmosphere(0)
-                current_level = Atmosphere(self._altitude.flatten(), altitude_in_feet=False)
-                for idx in range(len(self._true_airspeed)):
-                    delta = current_level.pressure[idx] / sea_level.pressure
-                    gamma = sea_level.density / current_level.density[idx]
-                    speed_of_sound = current_level.speed_of_sound[idx]
-                    altitude = current_level.get_altitude(altitude_in_feet=True)[idx]
-                    coeff = [
-                        3.0
-                        / 640.0
-                        * (1.0 - 10.0 * delta + 9 * delta ** 2)
-                        * 1
-                        / speed_of_sound ** 4
-                        * gamma ** 2,
-                        0,
-                        1.0 / 8.0 * (1.0 - delta) * 1 / speed_of_sound ** 2 * gamma,
-                        0,
-                        1.0,
-                        -cas.flatten()[idx],
-                    ]
-                    real_positive_roots = np.array(
-                        [np.real(x) for x in np.roots(coeff) if np.isreal(x) and x > 0.0]
-                    )
-                    if len(real_positive_roots) > 1:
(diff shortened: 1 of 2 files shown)


src/fastoad/model_base/atmosphere.py
```diff
@@ -182,6 +183,23 @@ def true_airspeed(self) -> Union[float, Sequence[float]]:
...
+                        / 5.0
+                        + 1.0
+                    )
+                    ** 3.5
+                    - 1.0
+                )
+                total_pressure = current_level.pressure + impact_pressure
+                sigma_0 = total_pressure / current_level.pressure
+                gamma = 1.4
+                mach = (2.0 / (gamma - 1.0) * (sigma_0 ** ((gamma - 1.0) / gamma) - 1.0)) ** 0.5
```
christophe-david, line 201: Here your formula explicitly uses gamma, but in the formula above, you use `3.5`, which is &#947;/(&#947;-1). For sake of consistency, the `gamma` variable should be used everywhere, or nowhere. And for sake of efficiency, I think that nowhere is the best choice.

Therefore, I suggest:
```suggestion
                mach = (5.0 * (sigma_0 ** (1.0 / 3.5) - 1.0)) ** 0.5
```

christophe-david (changes requested):
Thank you for this proposal.

As you will see, I have some comments, but the most important one is that you implemented a conversion that is valid only for subsonic flows.

Indeed, our current applications are at most transonic.
Indeed, the TAS<->CAS conversion in supersonic flows is [trickier](https://en.wikipedia.org/wiki/Calibrated_airspeed) 
And indeed, handling these two cases while taking into account the possible array-shape of inputs needs attention.

But the Atmosphere class should aim at being a solid and "universal" basis, and personally, knowing that EAS is close enough to CAS for the subsonic domain (considering the needed accuracy for conceptual studies), I prefer having no CAS computation than having a half-baked one.

Yet, I do not doubt that CAS computation is needed to have a complete Atmosphere class, but this computation should be reliable even above Mach 1.0.

areysset:
The formula has been changed to take into account [supersonic flight](https://en.wikipedia.org/wiki/Equivalent_airspeed)!

christophe-david:
> The formula has been changed to take into account [supersonic flight](https://en.wikipedia.org/wiki/Equivalent_airspeed)!

Ok, so it looks like my suggestion was a bad one, since results get further from reference values.
We will have to use the hard way, then...

areysset:
The new formulas taken from [Ref](https://rosap.ntl.bts.gov/view/dot/8909) highligh lower relative errors to reference results.
Also, we are not sure at which point reference results from [URL](http://www.aerospaceweb.org/design/scripts/atmosphere/) are 100% correct (error = 0%)?

christophe-david:
Another point: given your last modifications, I think you may have not read all my initial comments about gamma, and the mach-related code lines, did you ?
I can also add another point I should have spotted earlier: you can replace all your occurences of `current_level` by `self`.

areysset:
> Another point: given your last modifications, I think you may have not read all my initial comments about gamma, and the mach-related code lines, did you ?
> I can also add another point I should have spotted earlier: you can replace all your occurences of `current_level` by `self`.

I did miss the comment about gamma, sorry for that. I'll remove it definitly.
I'm gonna replace current_level then...

christophe-david:
Version 0.2.0 of [StdAtm](https://pypi.org/project/stdatm/) proposes a computation of calibrated airspeed, valid at subsonic and supersonic speeds, (and also dynamic and impact pressures).

PR #373, opened 2021-07-27
Issue 372 optimization viewer modifying output file

### LINKED ISSUE #372: Optimization viewer should not write outputs
Optimization viewer allows to change the "value" fields, and will save these values to the output file when the "save" button is clicked.
There is no reason to allow the modification of output file by this mean.

"value" fields should be read-only and no writing to output file should be done.
Additionally, when no output file exists, the optimization viewer should show nothing, or something like `-`, in the "value" fields.

ScottDelbecq:
This PR fixes #372.
Now the `optimization_viewer` does not modify the output file when clicking the `Save` button.
Also, the `Value` cells have been disabled and thus cannot be modified anymore.
Finally, in the absence of an output file, the `optimization_viewer` will display `-` in the `Value` cells instead of default values found in the models.

### DIFF
--- a/src/fastoad/gui/optimization_viewer.py
+++ b/src/fastoad/gui/optimization_viewer.py
@@ -75,6 +75,9 @@ class OptimizationViewer:
         # The ui containing save and load buttons
         self._save_load_buttons = None
 
+        # True if in the absence of an output file
+        self._MISSING_OUTPUT_FILE = None
+
     def load(
         self, problem_configuration: FASTOADProblemConfigurator,
     ):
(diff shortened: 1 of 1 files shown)


christophe-david (commented):
Thanks a lot for this quick PR. I am Ok with the result.

Just a problem: in the tutorial notebook, when no output file is available, if I change the initial value of the design var, it is not saved. Though it is actually saved if there IS an output file.
I could not find the reason for this behavior...

christophe-david (approved):
I decide to approve these changes, despite the small bug I identified in my previous comment, that **we will have to fix later**.

It will allow the PR for compatibility with OpenMDAO 3.10 to happen after this branch is merged.

PR #383, opened 2021-09-20
Issue 363 taper ratio

### LINKED ISSUE #363: data:geometry:wing:taper_ratio does not comply with the standard definition of taper ratio
`data:geometry:wing:taper_ratio` is used as an input by the legacy geometry module to quantify the ratio between `data:geometry:wing:root:virtual_chord` and `data:geometry:wing:tip:chord`.
> Note: _As documented in variable descriptions, this virtual chord is the chord length at wing root if sweep angle of trailing edge of outer wing part was on the whole wing (no kink)_

We currently cannot get rid of this input, as it is the way the module build the wing geometry, but the variable should be renamed like `data:geometry:wing:virtual_taper_ratio`.
And `data:geometry:wing:taper_ratio` should be created as an output to provide the ratio between `data:geometry:wing:root:chord` and `data:geometry:wing:tip:chord`.

Martin-Delavenne:
This PR resolve #363.

The "data:geometry:wing:taper_ratio" variable used as an input so far does not correspond to the classical definition of wing taper ratio. Therefore, not to modify all the geometry definition of the wing, the input variable was changed to "data:geoemetry:wing:virtual_taper_ratio" and an ouput "data:geometry:wing:taper_ratio" was created with the proper definition.

### DIFF
--- a/src/fastoad/cmd/tests/data/short_inputs.xml
+++ b/src/fastoad/cmd/tests/data/short_inputs.xml
@@ -70,7 +70,7 @@
       <wing>
         <aspect_ratio is_input="True">9.48<!--wing aspect ratio--></aspect_ratio>
         <sweep_25 units="deg" is_input="True">25.0<!--sweep angle at 25% chord of wing--></sweep_25>
-        <taper_ratio is_input="True">0.38<!--taper ratio of wing--></taper_ratio>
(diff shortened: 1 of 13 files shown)


christophe-david (approved):
Ok with this PR.
Thank you

PR #395, opened 2021-11-04
Issue 394 list modules + Release 1.1.1

### LINKED ISSUE #394: Version 1.1.0 breaks command line "fastoad list_modules"
**Describe the bug**
In FAST-OAD 1.1.0, command line `fastoad list_modules` does not accept configuration file anymore.

**To Reproduce**
```bash
fastoad gen_conf foo.yml
fastoad list_modules foo.yml
```

The command list native modules of FAST-OAD, but issues this warning:
`WARNING : SKIPPED toto.yml: folder does not exist.`

**Environment**
FAST-OAD 1.1.0

ScottDelbecq:
This PR fixes #394 making list_modules functional again using CLI.
This is the opportunity to do a patch release 1.1.1 (https://github.com/fast-aircraft-design/FAST-OAD/releases/untagged-7eeb0813b780611d787d) to include this bug correction.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ Changelog
 Version 1.1.1
 =============
 - Bug fixes:
-    - Fixed usage of list_modules with CLI. (#394)
+    - Fixed usage of list_modules with CLI. (#395)
 
 Version 1.1.0
 =============
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 1.1.1
+=============
+- Bug fixes:
+    - Fixed usage of list_modules with CLI. (#394)
+
 Version 1.1.0
 =============
 - Changes:
--- a/src/fastoad/cmd/fast.py
+++ b/src/fastoad/cmd/fast.py
@@ -79,9 +79,13 @@ class Main:
     @staticmethod
     def _list_modules(args):
         """Prints list of system identifiers."""
-        api.list_modules(
-            args.source_path, out=args.out_file, overwrite=args.force, verbose=args.verbose
-        )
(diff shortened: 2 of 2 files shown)


src/fastoad/cmd/fast.py
```diff
@@ -266,6 +266,7 @@ def run(self):
             help="Provides the identifiers of available systems",
             description="Provides the identifiers of available systems",
         )
+        self._add_conf_file_argument(parser_list_modules, required=False)
```
christophe-david, line 269: This line has not to be put back, since the argument "source_path" is now defined below.
```suggestion
```

christophe-david (changes requested):
Did you tested this solution ? It does not work at all for me.
The problem is that `oad.list_modules()` is logically fed with a list, even if the CLI is supplied with one argument. And when supplied with a list, `oad.list_modules()` assumes all elements in the list are folders.
Therefore, 2 solutions:
- providing a string instead of a 1-element list to `oad.list_modules()` when only one argument is provided to CLI
- Considering that each element in a list provided to `oad.list_modules()` can also be a file

src/fastoad/cmd/fast.py
ScottDelbecq: Thanks. Corrected in 69af999.
I prefer solution 1 as `om.list_modules()` is working as expected.

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 1.1.1
+=============
+- Bug fixes:
+    - Fixed usage of list_modules with CLI. (#394)
```
christophe-david, line 8:
```suggestion
    - Fixed usage of list_modules with CLI. (#395)
```

christophe-david (approved):
Ok, thank you.
Just a minor change in changelog. I'll do it myself.

PR #405, opened 2021-12-14
CLI refactoring

christophe-david:
This PR reworks the code for command line interface using [click](https://click.palletsprojects.com/en/8.0.x/).

It makes the code much more simpler.


ScottDelbecq (approved):
Thanks for this PR! `click` looks very convenient.
Does it enable to add tests for `cli.py`?

christophe-david:
> Does it enable to add tests for `cli.py`?

Yes, `click` comes with a small API for testing that should ease the process. I did not want to take the time now to ad these tests, since we did not have tests for the previous solution, but it should definitely been done.
I'll raise an issue about it to keep that in mind.

PR #408, opened 2022-01-10
Issue 406 - Adding internal IndepVarComp to inputs

### LINKED ISSUE #406: Undetected inputs when using IndepVarComp in a problem
**Describe the bug**
When generating the inputs of a problem containing an `IndepVarComp`, the outputs of the `IndepVarComp` should be considered as inputs and hence present in the generated `.xml` file.

**To Reproduce**
Steps to reproduce the behavior. 
```python
import openmdao.api as om
from fastoad.api import generate_inputs

# build the model
prob = om.Problem()

# System variables
comp = om.IndepVarComp()
comp.add_output("system:x", val=3.0)
comp.add_output("system:y", val= -4.0)

prob.model.add_subsystem("indeps", comp, promotes=["*"])

prob.model.add_subsystem("paraboloid", om.ExecComp("f = (x-3)**2 + x*y + (y+4)**2 - 3"), promotes=["*"])

prob.model.connect("system:x", "x")
prob.model.connect("system:y", "y")
```

**Expected behavior**
`"system:x"` and `"system:y"` should appear in the input file which is not the case

**Error message**

**Environment**
- Operating System: Windows 10
- Python environment: Python 3.7.11>
- Packages: fast-oad==1.2.1 (master), openmdao==3.10.0

**Additional context**

ScottDelbecq:
This PR allows to include internal `IndepVarComp`s in problem inputs, hence modifiying their value.
Therefore, it closes #406.

For this purpose the generation of the input file uses the `VariableList.from_problem()` method instead of the function `get_unconnected_input_names`. Now `get_unconnected_input_names` is only used in unit tests and could be removed later.
To make this functional, it was necessary to change the manner that the initial values (and other metadata) of inputs were set.
Now we use the `problem.set_value()` method after the `problem.setup()`.


src/fastoad/io/configuration/configuration.py
```diff
@@ -414,33 +430,39 @@ def _add_design_vars(self, model, auto_scaling):
...
         Needed variables are returned as an IndepVarComp instance while unused variables are
         returned as a VariableList instance.
 
         :param problem: problem with missing inputs. setup() must have been run.
         :return: IVC of needed input variables, VariableList with unused variables.
         """
-        mandatory, optional = get_unconnected_input_names(problem, promoted_names=True)
-        needed_variable_names = mandatory + optional
+        # TODO: shift this to from_problem
+        problem_copy = copy.deepcopy(problem)
```
christophe-david, line 446: You warned me about this and indeed, this is strange that we need that. Even stranger is the fact that it breaks one (and only one) integration test....

src/fastoad/io/configuration/configuration.py
```diff
@@ -414,33 +430,39 @@ def _add_design_vars(self, model, auto_scaling):
...
+        problem_variables = VariableList().from_problem(problem_copy)
+        problem_inputs_names = [var.name for var in problem_variables if var.is_input]
```
christophe-david, line 448: Here also, this line will be avoided if `from_problem` does the job of providing only inputs.

tests/integration_tests/oad_process/test_oad_process.py
```diff
@@ -61,6 +61,7 @@ def test_oad_process(cleanup):
     # Create problems with inputs
     problem = configurator.get_problem(read_inputs=True)
     problem.setup()
+    configurator.set_initial_values(problem)
```
christophe-david, line 64: Here also, should be removed if my suggestion are applied.

tests/integration_tests/oad_process/test_oad_process.py
```diff
@@ -164,6 +165,7 @@ def run_non_regression_test(
     # Get problem with inputs -------------------------------------
     problem = configurator.get_problem(read_inputs=True)
     problem.setup()
+    configurator.set_initial_values(problem)
```
christophe-david, line 168: Here also, should be removed if my suggestion are applied.

christophe-david (changes requested):
Sorry, I think this PR has to be deeply reworked, because it raises several architecture issues:

The first thing that alerted me is that the only way you could find to test your modifications was to create configuration files. If you think about the needs for the PR here they are:
  - Obtaining the list of inputs from a problem, including indep vars. It happens that `VariableList.from_problem()` does the job. We just miss a test that shows this feature.
  - Being able to read problem inputs in a way that allows to redefine indep vars, i.e. using `om.Problem.set_val`, and that needs to be done after setup. And to me, that belongs in the `FASTOADProblem` class.  
  The previous solution, i.e. inserting an IndepVarComp for inputs in the model, needed to do the input reading when creating the problem, and then the logical choice was to do it in the `FASTOADProblemConfigurator` method that creates the problem. But in terms of API, that was already something weird to have the capability of writing outputs in `FASTOADProblem` class, and not the ability to read inputs.  
  But not only does the new solution allows reading inputs from the `FASTOADProblem` class, it is a serious incentive to do that. The best proof for me is that in your tests, you create the problem, do the setup and then, you have to call the `FASTOADProblemConfigurator` method and provide the created problem to it. That is a weird flow.

In conclusion, the needed features should not be in `FASTOADProblemConfigurator`, and the new unit tests should not concern it.

That being said, such changes would modify the API and have to be managed carefully.

Therefore, I suggest following plan:
  1. Adding a unit test for `VariableList.from_problem()` that shows that indep vars are provided as inputs.
  2. Modifying `VariableList.from_problem()` to add a new argument `io_status` that could be `"inputs"`, `"outputs"` or `"all"` (default value). Let's be bold and have no check on the provided value for this argument. And the unit tests for this argument have to be very simple.
  3. Moving the `set_initial_values()` and `_get_problem_inputs` methods to `FASTOADProblem` class. Making `set_initial_values()` private (see below why) by renaming it to `_read_inputs`. And ensuring that it will feed `FASTOADProblem.additional_variables`.
  4. Adding a boolean flag `_read_inputs_after_setup` to `FASTOADProblem` class (set to `False` at instantiation)
  5. Creating a `read_inputs` method in `FASTOADProblem` class that will do this:
     - If setup has been run, simply run `set_initial_values()`
     - If setup has not been run, set the `_read_inputs` flag to True
  6. Overloading `setup()` method to call `_read_inputs()` after the original setup if `_read_inputs_after_setup` is `True`
  7. Modifying `FASTOADProblemConfigurator` so it uses `FASTOADProblem` methods. In particular, `FASTOADProblemConfigurator.get_problem()` should call `FASTOADProblem.read_inputs()` if `read_inputs` is True.
  8. Removing the unit tests that were added for `FASTOADProblemConfigurator`.

src/fastoad/openmdao/variables.py
```diff
@@ -699,9 +705,21 @@ def from_problem(
...
+        elif io_status == "outputs":
+            variables = output_vars
+        else:
+            raise TypeError("Unknown value for io_status")
+
+        return variables
 
     @classmethod
+    @deprecated(
+        version="1.3.0", reason="Will be removed in version 2.0. Please use from_problem() instead"
```
christophe-david, line 721:
```suggestion
        version="1.3.0", reason="Will be removed in version 2.0. Please use VariableList.from_problem() instead"
```

christophe-david (changes requested):
Thank you for this big update. There still some details that I think should be addressed.

src/fastoad/openmdao/problem.py
```diff
@@ -63,3 +83,48 @@ def write_outputs(self):
...
+    def read_inputs(self):
+        """
+        Reads inputs of the problem.
+        """
+        if self._metadata is not None and self._metadata["setup_status"] == _SetupStatus.POST_SETUP:
+            self._read_inputs()
+        else:
+            self._read_inputs_after_setup = True
+
+    def _get_problem_inputs(self) -> Tuple[om.IndepVarComp, VariableList]:
```
christophe-david, line 96: PyCharm helped me to spot this one:
```suggestion
    def _get_problem_inputs(self) -> Tuple[VariableList, VariableList]:
```

christophe-david (approved):
Thank you. I'm Ok with this PR. Just one tiny fix to apply and it can be merged.

christophe-david:
Don't bother. Working on a solution...

Issue #407, opened 2021-12-17
Testing CLI


christophe-david:
FAST-OAD uses [`click`](https://click.palletsprojects.com/en/8.0.x/) since PR #405. It should now be possible to build tests for the CLI, which should be done.

PR #409, opened 2022-01-13
plugin complement

### LINKED ISSUE #393: Custom configuration files could be associated to custom modules
FAST-OAD currently does not propose a way to provide a sample configuration file adapted to custom models.

Developers of custom modules should have a way to add a sample configuration file to the provided models. The FAST-OAD command for generating a configuration file should propose a way to choose the desired sample configuration file.

christophe-david:
This PR implements an enhanced plugin system and closes #393.

It contains following modifications:
- Documentation has been updated. Preview is [here](https://fast-oad.readthedocs.io/en/issue-393_plugin-complement/).
- Plugin declaration now uses `fastoad.plugins` as group name instead of `fastoad_model`
  - The previous group name `fastoad_model` is still usable, though not mentioned anymore in RTD documentation. It works the same as before, i.e. it expects to have only FAST-OAD modules in the declared plugin package.
  - When using the new group name, configuration files and notebooks can be part of the plugin, and a certain folder structure is expected in the plugin package (see [doc](https://fast-oad.readthedocs.io/en/issue-393_plugin-complement/documentation/custom_modules/add_plugin.html)).
- The implemented modifications make that models are loaded only at first instantiation of the new class `FastoadLoader`. Then the loading is not made anymore at startup, which greatly speeds up commands like `fastoad -h` or `fastoad -v`.
- CLI and API have been updated (see [doc](https://fast-oad.readthedocs.io/en/issue-393_plugin-complement/documentation/usage.html#using-fast-oad-through-command-line)):
  - modified configuration file generation to allow to choose which plugin to use to generate a configuration file
  - modified notebook creation to possibly limit the generated notebooks to the one from a specified plugin. Also, `fastoad.api` now contains a `create_notebooks()` method.
  - added a command to get plugin information.
- Modified and added commands in CLI are now unit-tested. It starts the resolution of #407.
- The test of the plugin mechanism has been made independent of actual declared plugins using fixtures in `contest.py`. It has been rather painful to achieve that 😅.
- Usage of `pkg_resources` has been removed to use [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html) as it is now recommended.
- `tests/testing_utilities.py` has been moved to `src/fastoad/_utils/testing.py` so that there shouldn't be anymore warning when importing copied models in other projects.


docs/documentation/custom_modules/add_plugin.rst
```diff
@@ -60,10 +150,29 @@ Let's assume your :code:`pyproject.toml` file is configured so that your project
...
 
-It will create a :code:`dist` folder with two files: :code:`ST_drive_models-1.0.0.tar.gz`
-and :code:`ST_drive_models-1.0.0-py3-none-any.whl` (or something like this).
+The command :code:`poetry build` will create a :code:`dist` folder with two files:
+
+:code:`ST_drive_models-1.0.0.tar.gz` and :code:`ST_drive_models-1.0.0-py3-none-any.whl`
+(or something like this).
 
 You may then have sent any of those two files to another user, who may then install your models
 using :code:`pip` with:
```
ScottDelbecq, line 178: In the section "Publishing" I would underline that to publish on PyPi the package name will be the one provided in the field `name` in the `pyproject.toml`. Therefore, the publisher shall check that the chosen package name is not already taken on PyPi.

ScottDelbecq (approved):
Thank you for this PR!
This is really what I had in mind for the plugin features.
I have a small suggestion for the documentation (c.f. my only comment).

docs/documentation/custom_modules/add_plugin.rst
christophe-david: Agreed. Done in a64d130b46db8f1102f014668d822f0b528c5921.

PR #433, opened 2022-04-07
Added nbstripout to dependencies and pre-commit

ScottDelbecq:
This small PR adds [nbstripout](https://github.com/kynan/nbstripout) to dependencies and pre-commit hooks.
This way notebook outputs are deleted when notebooks are committed. 
I tested the pre-commit and it seems to work fine.
Once approved and merged I will do the same modifications for the [CS_25](https://github.com/fast-aircraft-design/FAST-OAD_CS25) repository which contains more notebooks.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,3 +20,10 @@ repos:
         language: python
         pass_filenames: false
         files: 'poetry.lock'
+
+      - id: nbstripout
+        name: nbstripout
+        description: 'nbstripout: strip output from Jupyter and IPython notebooks'
+        entry: nbstripout
+        language: python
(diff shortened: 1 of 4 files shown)


christophe-david (approved):
That's a good addition, that will have to be copied in FAST-OAD-CS25 :)

Just a question: why not having used [the way proposed by the author](https://github.com/kynan/nbstripout#using-nbstripout-as-a-pre-commit-hook)? Besides the fact that I would generally trust the author about how his tool should be used, the proposed way has the (very) small avantage of not having to add the package in Poetry.

ScottDelbecq:
I agree but I was willing to have it in the dev dependencies to use it whenever I want without waiting for a commit :)

PR #435, opened 2022-05-13
Datafile enhancement

### LINKED ISSUE #421: DataFile.from_*() methods return a VariableList instance
**Describe the bug**
Methods like DataFile.from_dataframe() or DataFile.from_ivc() return a VariableList instance instead of a DataFile instance.

**To Reproduce**
```python
import fastoad.api as oad
import openmdao.api as om

ivc = om.IndepVarComp("my_var")
result = oad.DataFile.from_ivc(ivc)
print(type(result))
```
returns
```
<class 'fastoad.openmdao.variables.variable_list.VariableList'>
```

**Expected behavior**
The snippet above should return `<class 'fastoad.io.variable_io.DataFile'>`. More generally, any subclass of `VariableList` should return an instance of the subclass when using this methods.

### LINKED ISSUE #429: DataFile class should report an error when opening a non existent file
**Describe the bug**
When using DataFile to read a non existing file, no error is triggered and the obtained instance behave as if the file was empty.

**To Reproduce**
```python
>>> import fastoad.api as oad
>>> my_vars = oad.DataFile("/non/existent/file")
>>> my_vars.names()
[]
```

**Expected behavior**
The idea behind this behavior is to provide the ability to create a DataFile for a non-existing file, to populate it with variables and then to write it.
But the drawback is that it does not warn when the user wants to read an existing file and just does a typo.

Maybe a class method `DataFile.new(path)` would allow to create a new instance with provide path, while the classic instantiation would require an existing file.

christophe-david:
This PR aims at easing a bit the usage of DataFile class:
- fixes #421 and #429
- allows instantiating a DataFile with no argument or from list of a Variable instances
- adds the `save_as()` method
- adds a documentation section about usage of DataFile (preview [here](https://fast-oad.readthedocs.io/en/datafile-enhancement/documentation/variables.html#fast-oad-api))

BTW, I changed a bit the GitHub workflows to take advantage of the cache management of the `setup-python` action, that works for Poetry (bye bye `actions/cache`). I used this branch as a sandbox to test it.

Finally, since CodeClimate was nagging me with a "two many nested control flows" in the dreaded `VariableList.from_problem()` method (just because I slightly modified it to fix issue 421), I decided to comply by simply using a [guard clause](https://en.wikipedia.org/wiki/Guard_(computer_science)).

### DIFF
--- a/docs/documentation/data.xml
+++ b/docs/documentation/data.xml
@@ -1,0 +1,17 @@
+<FASTOAD_model>
+  <data>
+    <geometry>
+      <wing>
+        <area units="m**2">150.0</area>
+      </wing>
+    </geometry>
+    <weight>
+      <fuselage>
+        <mass units="kg">10000.0</mass>
+        <CG>
+          <x units="m">20.0</x>
+        </CG>
+      </fuselage>
+    </weight>
+  </data>
+</FASTOAD_model>
--- a/docs/documentation/variables.rst
+++ b/docs/documentation/variables.rst
@@ -50,7 +50,24 @@ File format
 
 For writing input and output files, FAST-OAD relies on the path in the variable names.
 
-For example, for the three variables above, the matching XML file will be:
+For instance, the variable name :code:`data:geometry:wing:area` will be split according
+to colons :code:`:` and each part of the name will become a level in the XML hierarchy:
+
+.. code-block:: xml
+
+        <data>
+            <geometry>
+                <wing>
+                    <area units="m**2">
+                        150.0
+                    </area>
+                </wing>
+            </geometry>
+        </data>
+
+
+A complete file that would contain the three above-mentioned variables will be as following:
+
 
 .. code-block:: xml
 
@@ -76,6 +93,12 @@ For example, for the three variables above, the matching XML file will be:
 
     Units are given as a string according to
     `OpenMDAO units definitions <http://openmdao.org/twodocs/versions/latest/features/units.html>`_
+
+.. note::
+
+    XML requires a unique root element for containing all other ones. Its name can be
+    freely chose, but it is `FASTOAD_model` in files written by FAST-OAD
+
 
 FAST-OAD API
 ************
@@ -126,6 +149,28 @@ Provided that above file is named :code:`data.xml`, following commands apply:
     >>> datafile.file_path  # The object is now associated to the new path
     './new_data.xml'
 
+After running these lines of code, the generated file :code:`new_data.xml` contains:
+
+.. code-block:: xml
+
+    <FASTOAD_model>
+        <data>
+            <geometry>
+                <fuselage>
+                    <length units="m">35.0</length>
+                </fuselage>
+                <wing>
+                    <area units="m**2">120.0</area>
+                    <mass units="kg">10500.0</mass>
+                </wing>
+            </geometry>
+            <weight>
+                <fuselage>
+                    <mass units="kg">10000.0</mass>
+                </fuselage>
+            </weight>
+        </data>
+    </FASTOAD_model>
 
 
 
--- a/docs/documentation/variables.rst
+++ b/docs/documentation/variables.rst
@@ -17,6 +17,9 @@ available for the whole process.
 The list of variable names and descriptions for a given problem can be obtained from command line (see
 :ref:`get-variable-list`).
 
+.. contents::
+   :local:
+
 
 ***************
 Variable naming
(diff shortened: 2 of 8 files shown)


docs/documentation/variables.rst
```diff
@@ -42,28 +45,89 @@ The other path elements depend of the variable. The number of path elements is n
...
+            <weight>
+                <fuselage>
+                    <mass units="kg">10000.0</mass>
+                    <CG>
+                        <x units="m">20.0</x>
+                    </CG>
+                </fuselage>
+            </weight>
+        </data>
+    </FASTOAD_model>
```
florentLutz, line 73: Maybe it is obvious, but shouldn't we say that each : corresponds to an indent in the input and output files ?

docs/documentation/variables.rst
```diff
@@ -42,28 +45,89 @@ The other path elements depend of the variable. The number of path elements is n
...
+    ['data:geometry:wing:area', 'data:weight:fuselage:mass', 'data:geometry:fuselage:length', 'data:geometry:wing:mass']
+    >>> # ---------------------------------
+    >>> # Writing to another file
+    >>> datafile.save_as("./new_data.xml", overwrite=True)
+    >>> datafile.file_path  # The object is now associated to the new path
+    './new_data.xml'
+
+
+
+
```
florentLutz, line 131: Maybe do a similar thing as for data.xml and show that new_data.xml indeed corresponds to what we expect like so :
 ```suggestion
 .. code-block:: xml
    <FASTOAD_model>
        <data>
            <geometry>
                <fuselage>
                    <length units="m">35.0</mass>
                </fuselage>
                <wing>
                    <area units="m**2">150.0</area>
                </wing>
            </geometry>
            <weight>
                <fuselage>
                    <mass units="kg">10000.0</mass>
                </fuselage>
                <wing>
                    <mass units="kg">10500.0</mass>
                </wing>
            </weight>
        </data>
    </FASTOAD_model>
```

File: docs/documentation/data.xml [hunk 3]
```diff
@@ -0,0 +1,17 @@
...
+    <weight>
+      <fuselage>
+        <mass units="kg">10000.0</mass>
+        <CG>
+          <x units="m">20.0</x>
+        </CG>
+      </fuselage>
+    </weight>
+  </data>
+</FASTOAD_model>
```
florentLutz, line 17: Is this file necessary ? Since in the documentation it is shown as a code-block and not as reading this file (which I don't know if it is possible)

src/fastoad/io/tests/test_data_file.py
```diff
@@ -46,18 +46,39 @@ def write_variables(self, data_source: Union[str, IO], variables: VariableList):
...
-def test(cleanup):
+def test_DataFile(cleanup):
+    variables_ref = VariableList([Variable("data:foo", value=5), Variable("data:bar", value=10)])
+
     file_path = pth.join(RESULTS_FOLDER_PATH, "dummy_data_file.xml")
-    variables_1 = DataFile(file_path)
-    assert len(variables_1) == 0
+    with pytest.raises(FileNotFoundError) as exc_info:
+        _ = DataFile(file_path)
+    assert exc_info.value.args[0] == f'File "{file_path}" is unavailable for reading.'
```
florentLutz, line 55: I suggest adding a comment that explains what all the assertion do. On this one, which I assume raises an error when reading a non existant file could be 
```python
# Tests that an error is raised when opening non_existent file (Issue #429)
```

src/fastoad/io/variable_io.py
```diff
@@ -49,6 +50,10 @@ def read(self, only: List[str] = None, ignore: List[str] = None) -> VariableList
         :param ignore: List of variable names that should be ignored when reading.
         :return: an VariableList instance where outputs have been defined using provided source
         """
+        if isinstance(self.data_source, str) and not isfile(self.data_source):
+            raise FileNotFoundError(
+                f'File "{self.data_source}" is unavailable for reading.'
+            ) from FastError()
```
florentLutz, line 56: In some other files, when a line corrects an issue it is sometimes mentionned. Out of curiosity what's the "good practice" that decides when to fo it or not

florentLutz (commented):
A few comments on the style since I won't be able to fully judge the conten but it does not preclude the merge.

docs/documentation/variables.rst
christophe-david: Well, that has nothing to do with indents (we talk about XML), but I see your point. Indeed, I though that it was rather obvious, given the content of above section. You think it's not ?

src/fastoad/io/tests/test_data_file.py
christophe-david: You understood the purpose, which tends to show that the comment is not mandatory :)

Generally speaking, comments that just paraphrase the code are considered bad practice.
Comments should be used only to explain why a given solution has been chose, or to help understanding a particularly complex piece of code (which is generally the sign that the code has problems, but as mere humans, we sometimes have to cope with that).

Here, the only additional information I see would be the source bug. Though I am willing to mention the issue for very peculiar bugs, the checked feature here is very basic, so I do not see the need.

Do you agree?

src/fastoad/io/variable_io.py
christophe-david: Well, I answered just above :)
Yet, do not consider it as absolute truth. That is also my own interpretation.

docs/documentation/variables.rst
florentLutz: I can think of one student for which it wasn't that obvious. It may juste be an exception. We can keep it as is and eventually change it later based on user feedback.

src/fastoad/io/tests/test_data_file.py
florentLutz: Indeed, but because I read the issue beforehand.
Another question out of curiosity, is it assumed that basic users (as opposed to devs) won't go into the tests for informations on what is possible to do (meaning we can assume that the people who will read this have a better understanding of what is done) ? 
I wasn't sure that the good practice also applied to tests, but now I know :)
We can resolve this then.

docs/documentation/variables.rst
christophe-david: I tried something. Let me know if it looks Ok to you.

docs/documentation/variables.rst
christophe-david: Done in 93424f583d1759b2a32eebac09abe92a7c9ae9df

docs/documentation/variables.rst
florentLutz: The explanation looks good to me, and I think I get why the example on the wing area looks like this [value not on the same line as the name and unit of the variable (again maybe "line" is not the right term for XML)], but isn't it too different from the other examples where name, value and unit are all on the same line ?

florentLutz (approved):

florentLutz (approved):
Just had one last comment but it does not justify not merging the PR

docs/documentation/variables.rst
christophe-david: Actually, "line" is fine, since it applies about the file layout. But in terms of XML, the term is irrelevant because line breaks do not matter, so there is strictly no difference between the examples.
I wanted the first example to clearly show the path breakdown, which is why I added these additional line breaks. I really hope that anyone disturbed by the layout discrepancies between samples will have a second look and find out they are equivalent.

florentLutz: We mentionned it yesterday but we saifthat the users that were going to read the documentation were already a bit "fluent" in Python, so I guess it makes sense to assume that they would either know that line breaks don't matter in XML or have the hindsight to find out on their own.

PR #443, opened 2022-06-09
Memory fix

christophe-david:
OpenMDAO objects were persisting in RAM after setup or run operations.
This could be a problem when running multiple computations in the same Python session: memory consumption was continuously raising, and computation time was increasing.

This was due to iPOPO instantiating OpenMDAO components, but never releasing them, since we did not ask to.

This PR solves this problem in a somehow rough way, but it apparently works fine.

### DIFF
--- a/src/fastoad/openmdao/problem.py
+++ b/src/fastoad/openmdao/problem.py
@@ -28,6 +28,7 @@ from fastoad.openmdao.validity_checker import ValidityDomainChecker
 from fastoad.openmdao.variables import VariableList
 from ._utils import problem_without_mpi
 from .exceptions import FASTOpenMDAONanInInputFile
+from ..module_management._bundle_loader import BundleLoader
 
 # Name of IVC that will contain input values
 INPUT_SYSTEM_NAME = "fastoad_inputs"
@@ -63,11 +64,13 @@ class FASTOADProblem(om.Problem):
     def run_model(self, case_prefix=None, reset_iter_counts=True):
         status = super().run_model(case_prefix, reset_iter_counts)
         ValidityDomainChecker.check_problem_variables(self)
+        BundleLoader().clean_memory()
         return status
 
     def run_driver(self, case_prefix=None, reset_iter_counts=True):
         status = super().run_driver(case_prefix, reset_iter_counts)
         ValidityDomainChecker.check_problem_variables(self)
+        BundleLoader().clean_memory()
         return status
 
     def setup(self, *args, **kwargs):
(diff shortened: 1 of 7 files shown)


src/fastoad/openmdao/problem.py
```diff
@@ -63,11 +64,13 @@ def __init__(self, *args, **kwargs):
     def run_model(self, case_prefix=None, reset_iter_counts=True):
         status = super().run_model(case_prefix, reset_iter_counts)
         ValidityDomainChecker.check_problem_variables(self)
+        BundleLoader().clean_memory()
         return status
 
     def run_driver(self, case_prefix=None, reset_iter_counts=True):
         status = super().run_driver(case_prefix, reset_iter_counts)
         ValidityDomainChecker.check_problem_variables(self)
+        BundleLoader().clean_memory()
```
ScottDelbecq, line 73: If I comment the two lines (and leave `.clean_memory()` only after the `.setup()`) your test passes. 
If I change your test with a single setup and multiple `run_model`:

```python
def run_problem():
    configurator = oad.FASTOADProblemConfigurator(pth.join(DATA_FOLDER_PATH, "oad_process.yml"))
    configurator.input_file_path = pth.join(RESULTS_FOLDER_PATH, "inputs.xml")
    configurator.output_file_path = pth.join(RESULTS_FOLDER_PATH, "outputs.xml")
    print_memory_state("After reading configuration file")

    ref_inputs = pth.join(DATA_FOLDER_PATH, "CeRAS01.xml")
    configurator.write_needed_inputs(ref_inputs)
    print_memory_state("After writing input file")

    problem = configurator.get_problem(read_inputs=True)
    print_memory_state("After building problem")

    problem.setup()
    print_memory_state("After problem setup")

    count = 100
    for i in range(count):
        print(f"RUN {i+1}/{count}")
        problem.run_model()

    print_memory_state("After problem run")

    problem.write_outputs()
```

The test still passes. Therefore, I wonder if it is necessary to clean the memory after a `run_model`or `run_driver`. Or our tests are not stringent enough? What do you think?

ScottDelbecq (changes requested):
Thank you for the PR!

I have a single comment that discusses whether memory cleaning is necessary after running a model. Nevertheless, it can be seen as us being cautious without impacting performance.

Also, it seems that the new memory test is not integrated in the CI yet.

christophe-david:
> Also, it seems that the new memory test is not integrated in the CI yet.

Done in cfcb69f805ff52fa3b01448f31b9a4b945c2c0a0

ScottDelbecq (approved):

Issue #440, opened 2022-05-24
No reset of active_submodel when pytest are run at the same time


florentLutz:
**Describe the bug**
Running two OAD processes in two pytests in the same pytest file will not reset to the default submodels from one process to another. In the small code provided in next step, running both pytests at the same time (e.g. running pytest submodel_bug/test_submodel.py
) will provide an error because in the second pytest the default submodel is not used, it is put to null as is the case in the first test. When running both tests separetely however, there is non error

**To Reproduce**
See little code attached : 
[submodel_bug.zip](https://github.com/fast-aircraft-design/FAST-OAD/files/8764007/submodel_bug.zip)

**Expected behavior**
Expected behaviour of when tests are run one right after the other is the behaviour when both tests are run separately

**Environment**
- Operating System: Windows 10 Entreprise
- Python environment: Python 3.9.11
- Packages:
aenum==3.1.8
anyio==3.5.0
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
asttokens==2.0.5
atomicwrites==1.4.0
attrs==21.4.0
Babel==2.9.1
backcall==0.2.0
beautifulsoup4==4.10.0
black==22.3.0
bleach==4.1.0
certifi==2021.10.8
cffi==1.15.0
charset-normalizer==2.0.12
click==8.1.2
colorama==0.4.4
debugpy==1.6.0
decorator==5.1.1
defusedxml==0.7.1
Deprecated==1.2.13
ensure==1.0.2
entrypoints==0.4
executing==0.8.3
fast-oad-core==1.3.3
fastjsonschema==2.15.3
idna==3.3
importlib-metadata==4.11.3
iniconfig==1.1.1
iPOPO==1.0.1
ipykernel==6.12.1
ipysheet==0.5.0
ipython==8.2.0
ipython-genutils==0.2.0
ipywidgets==7.7.0
jedi==0.18.1
Jinja2==3.1.1
json5==0.9.6
jsonrpclib-pelix==0.4.3.2
jsonschema==4.4.0
jupyter-client==7.2.1
jupyter-core==4.9.2
jupyter-server==1.16.0
jupyterlab==3.3.2
jupyterlab-pygments==0.1.2
jupyterlab-server==2.12.0
jupyterlab-widgets==1.1.0
lxml==4.8.0
MarkupSafe==2.1.1
matplotlib-inline==0.1.3
mistune==0.8.4
mypy-extensions==0.4.3
[...]

PR #444, opened 2022-06-10
Issue 440 reset null submodels

christophe-david:
Solves #440.

This PR ensures that all sub-models are reactivated before processing sub-models from the configuration file.

Some modifications have also been done to the [documentation](https://fast-oad.readthedocs.io/en/issue-440_reset-null-submodels/documentation/custom_modules/add_submodels.html).

### DIFF
--- a/docs/documentation/custom_modules/add_submodels.rst
+++ b/docs/documentation/custom_modules/add_submodels.rst
@@ -144,46 +144,26 @@ At this point, there are now 2 available submodels for the "atom_counter.wing" r
 do nothing else, the command :code:`oad.RegisterSubmodel.get_submodel("atom_counter.wing")` will
 raise an error because FAST-OAD needs to be instructed what submodel to use.
 
-The first way to do that is by Python. You may insert the following line at module level (i.e. not in
-any class or function):
+*****************************************
+How to specify which submodel to be used
+*****************************************
 
-.. code-block:: python
+There are two ways to specify which submodel has to be used:
 
-    oad.RegisterSubmodel.active_models["atom_counter.wing"] = "alternate.counter.wing"
+.. contents::
+   :local:
 
-The best place for such line would probably be in the module that defines your submodel. In this
-case, our above example would become:
+.. _submodel-spec-conf-file:
 
-.. code-block:: python
+Through configuration file (recommended)
+********************************************************
 
-    import openmdao.api as om
-    import fastoad.api as oad
-
-    oad.RegisterSubmodel.active_models["atom_counter.wing"] = "alternate.counter.wing"
-
-    @oad.RegisterSubmodel("atom_counter.wing", "alternate.counter.wing")
-    class CountWingAtoms(om.ExplicitComponent):
-        """Put another implementation here"""
-
-.. warning::
-
-    In case several Python modules define their own chosen submodel for the same requirement, the
-    last interpreted line will preempt, which is not a reliable way to do.
-    We currently expect such situation to be rare, where more than one alternate submodel would be
-    available (for the same requirement) in one set of FAST-OAD modules.
-    Anyway, in such situation, the only reliable way will be to use the configuration file, as
-    instructed below.
-
-**********************************************
-How to use submodels from configuration file ?
-**********************************************
-
-The second way to define what submodels should be used is by using FAST-OAD configuration file.
+The recommended way to specify submodels to be used is by using FAST-OAD configuration file.
 
 .. note::
 
     When it comes to the specification of submodels to be used, the configuration file will have
-    the priority over any Python instruction.
+    the priority over :ref:`Python instructions<submodel-spec-python>`.
 
 The configuration file can be populated with a specific section that will state the submodels
 that should be chosen.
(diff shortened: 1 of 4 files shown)


src/fastoad/module_management/service_registry.py
```diff
@@ -486,3 +486,10 @@ def get_submodel(cls, service_id: str, options: dict = None):
             instance = om.Group()
 
         return instance
+
+    @classmethod
+    def cancel_submodel_deactivations(cls):
+        """Reactivates all submodels that have been deactivated."""
+        deactivated_submodel_ids = [k for k, v in cls.active_models.items() if not v]
+        for submodel_id in deactivated_submodel_ids:
+            del cls.active_models[submodel_id]
```
florentLutz, line 495: Out of curiosity, why does deleting the deactivated submodels reactivate them ?

florentLutz (commented):
First of all thanks for the PR, I've tried it on the test code that was sent with the original issue and it works !

This has however raised a question for me :
- Let's assume that we had, in the file `dummy_submodels\submodel.py`, these two lines:
``` python
RegisterSubmodel.active_models["requirement.1"] = "req.1.submodel"
RegisterSubmodel.active_models["requirement.2"] = "req.2.submodel.A"
```
The same error as in the test will happen. I assumed, before re-reading the readthedocs that, since the second line was there, after the deactivation and deactivation cancel req.2.submodel.A would be used, which is not the case. It is mentioned in thereadthe docs that the last line preempts and that the configuration file is the right way to do it, but maybe this case should be highlighted, or at least the fact that deactivating a submodel counts as a choice of submodel selection.

src/fastoad/module_management/service_registry.py
christophe-david: The `RegisterSubmodel.active_models` class attribute is by default an empty dict. When a submodel is required, `RegisterSubmodel.active_models` is checked for the requirement ID. If it is found, then a choice has been expressed and the dict value is used (a submodel ID or `null`). If it is not found, then it means that no choice has been made, so either there is only one matching submodel, or an error will be raised.

Therefore, by removing the submodel ID from `RegisterSubmodel.active_models` keys, we simply get back to the no-choice state.

src/fastoad/module_management/tests/test_register_submodel.py
```diff
@@ -30,7 +30,11 @@
 @pytest.fixture(scope="module")
 def load():
     """Loads components"""
+    previous_active_submodels = RegisterSubmodel.active_models
+    RegisterSubmodel.active_models = {}
     BundleLoader().explore_folder(pth.join(DATA_FOLDER_PATH, "dummy_submodels"))
+    yield
+    RegisterSubmodel.active_models = previous_active_submodels
```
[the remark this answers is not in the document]
christophe-david: Yep. In a pytest fixture, `yield` is the point where the test that used the fixture is run, which allows to do stuff before and after the test.

For the record, the main usage of `yield` is when writing [generators](https://wiki.python.org/moin/Generators).

florentLutz (approved):

docs/documentation/custom_modules/add_submodels.rst
```diff
@@ -197,24 +177,42 @@ that should be chosen.
 In the above example, an alternate submodel is chosen for the "atom_counter.wing" requirement,
 whereas the original submodel is chosen for the "original.counter.fuselage" requirement (whether
 there is another one defined or not).
```
florentLutz, line 179: Is requirement equivalent to service here ?

docs/documentation/custom_modules/add_submodels.rst
```diff
@@ -144,46 +144,26 @@ At this point, there are now 2 available submodels for the "atom_counter.wing" r
...
-
-.. code-block:: python
-
-    import openmdao.api as om
-    import fastoad.api as oad
-
-    oad.RegisterSubmodel.active_models["atom_counter.wing"] = "alternate.counter.wing"
+*****************************************
+How to specify which submodel to be used
+*****************************************
```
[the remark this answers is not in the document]
christophe-david: I finally decided to rephrase entirely (6c295b7af765281a1fa198f144b3974097f2dcb7).
What do you think ?

docs/documentation/custom_modules/add_submodels.rst
```diff
@@ -144,46 +144,26 @@ At this point, there are now 2 available submodels for the "atom_counter.wing" r
...
+.. _submodel-spec-conf-file:
 
-**********************************************
-How to use submodels from configuration file ?
-**********************************************
+Through configuration file (recommended)
+********************************************************
 
-The second way to define what submodels should be used is by using FAST-OAD configuration file.
+The recommended way to specify submodels to be used is by using FAST-OAD configuration file.
```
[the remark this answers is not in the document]
christophe-david: Also rephrased

docs/documentation/custom_modules/add_submodels.rst
```diff
@@ -197,24 +177,42 @@ that should be chosen.
 In the above example, an alternate submodel is chosen for the "atom_counter.wing" requirement,
 whereas the original submodel is chosen for the "original.counter.fuselage" requirement (whether
 there is another one defined or not).
-No submodel is defined for the "atom_counter.empennage" requirement, which lets the choice to
-be done in Python, as explained in above sections.
+No submodel is defined for the "atom_counter.empennage" requirement. It will be Ok if only one
+submodel is available for this requirement. Otherwise, an error will be raised, unless the submodel
+choice is done through Python (see below).
```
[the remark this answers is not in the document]
christophe-david: 🆗 😄

docs/documentation/custom_modules/add_submodels.rst
```diff
@@ -110,18 +110,20 @@ To allow a simpler replacement of your submodels, you will need to use the
 This has the same behavior as the previous one, but the second one will allow substitution of
 submodels, as shown in next part.
 
-In details, :code:`CountWingAtoms` is declared as a submodel that fulfills the role of "wing atom
+In details, :code:`CountWingAtoms`is declared as a submodel that fulfills the role of "wing atom
```
florentLutz, line 113: Why deleting the space after :code:`CountWingAtoms` here and not after :code:`CountAtoms` below for instance ?

florentLutz (approved):
Selected submodel works for me :smile:

docs/documentation/custom_modules/add_submodels.rst
christophe-david: Fixed in ae8a74c5196861503dc1ea7971f1a0e4da41eb08

Issue #448, opened 2022-07-21
Unwanted generation of report folders.


christophe-david:
**Describe the bug**
Not exactly a bug, but with OpenMDAO 3.17+, each run may create a `reports` folder with some html files in it. Though this can be useful in some cases, having this folder every time is not desired.

A workaround is currently to set the environment variable `OPENMDAO_REPORTS` to `none` (see [OpenMDAO doc](https://openmdao.org/newdocs/versions/latest/features/model_visualization/reports_system.html#controlling-reporting-using-an-environment-variable))

**Expected behavior**
Getting back to previous behavior where no report was generated.
Possibly, an option in configuration file may allow to control this report behavior.

PR #449, opened 2022-07-21
Deactivated automatic reports.

christophe-david:
Solves #448.
Quick fix to avoid uncontrolled generation of `reports` folder with OpenMDAO 3.17+.

Solution is to deactivate entirely this feature, unless environment variable OPENMDAO_REPORTS is set. I suggest we see later if we add a control for it in the configuration file.

No test is implemented for this modification, since I prefer to keep our Poetry environment with the old OpenMDAO 3.10 (yet I tested it "manually", of course)). At least, watchman tests will ensure the change does not cause any crash.

### DIFF
--- a/src/fastoad/openmdao/problem.py
+++ b/src/fastoad/openmdao/problem.py
@@ -38,7 +38,8 @@ SHAPER_SYSTEM_NAME = "fastoad_shaper"
 
 
 class FASTOADProblem(om.Problem):
-    """Vanilla OpenMDAO Problem except that it can write its outputs to a file.
+    """
+    Vanilla OpenMDAO Problem except that it can write its outputs to a file.
 
     It also runs :class:`~fastoad.openmdao.validity_checker.ValidityDomainChecker`
     after each :meth:`run_model` or :meth:`run_driver`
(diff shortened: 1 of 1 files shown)


src/fastoad/openmdao/problem.py
```diff
@@ -38,14 +38,24 @@
...
 
     def __init__(self, *args, **kwargs):
+        if (
+            version.parse(openmdao.__version__) >= version.parse("3.17")
+            and "OPENMDAO_REPORTS" not in os.environ
+            and "reports" not in kwargs
+            and len(args) < 5
+        ):
+            # Automatic reports are deactivated for FAST-OAD, unless OPENMDAO_REPORTS env
+            # variable is set.
```
ScottDelbecq, line 57: Would it not be useful to handle this option via kwargs "reports" that could be specified in the configuration file or via `_run_problem(reports=True)? Maybe later?

ScottDelbecq (commented):
Thanks for the PR!
I know it is quite urgent to deactivate automatic reports, but I was wondering if it would be useful to handle this via `_run_problem(report=True)`(my preferred way)? And if yes should it be done in this PR or later?

src/fastoad/openmdao/problem.py
ScottDelbecq: Perfect! Then lets stick to this for the moment.

PR #453, opened 2022-08-18
Upgrade to wop 2.x API

relf:
This PR to remove [deprecated `login` parameter](https://github.com/OneraHub/WhatsOpt-CLI/releases/tag/2.0.0) from `WhatsOpt()` call. It was kept in wop 2.x so far not to break XDSM view feature here but time to move on. `wop` required version is also bumped to the latest (2.2.0).

Edit: I've also modified the source parameter of the `generate_xdsm_html()` call which is merely a string displayed in the footer of generated HTML page: "XDSM generated from <source>". Previously an empty string was rendered, now FAST-OAD x.y.z is displayed.


christophe-david (approved):
Thanks for the contribution.

We just miss the update of the poetry.lock file( and of the docs/requirements.txt, done by pre-commit hook when poetry.lock is modified), but I'll do that after the merge.

PR #456, opened 2022-09-27
Issue 455 variable viewer fix

### LINKED ISSUE #455: Compatibility of variable viewers with "simple" variable names
**Describe the bug**
Variable viewer behaves poorly when variables names do not contain any colon (`:`) separator

**To Reproduce**
When no colon separator is used in variable names of a given problem, the variable viewer is unable to show the complete list of variable in the table.
All variables are accessible through the drop-down filter menu, but onlu one variable will appear in the table.

image

**Expected behavior**
The first drop-down menu misses a `ALL` value that should be selected by default.

ScottDelbecq:
This PR resolves #455 by adding the `--ALL--` tag for the first filter of the variable viewer. This way, by default, the variable viewer will render all the variable of the system.

Also, due to incompatibility between ipywidgets and jupyterlab versions in the .lock file (for some OS such as OSX), the minimal versions have been updated in the pyproject.toml and thus the .lock.


src/fastoad/gui/variable_viewer.py
```diff
@@ -423,7 +425,7 @@ def _filter_variables(
             else:
                 path = ":".join(modules)
 
-        path_filter = [True] * len(df) if path == "" else df.Name.str.startswith(path)
+        path_filter = [True] * len(df) if path == ("" or TAG_ALL) else df.Name.str.startswith(path)
```
christophe-david, line 428: At first, you made me think that if `path == ("" or TAG_ALL)` could be a Python feature I did not know about. But in my tests, it does not work: `"" or TAG_ALL` is actually `TAG_ALL` so your test is equivalent to `path == TAG_ALL`.
If you want to test that, you need to write `path in ["", TAG_ALL]`. But if you do so, then the line `filtered_df = df[path_filter & io_filter]` below may fail because `&` does not work if given 2 lists (`[True] * len(df)`) 

And it looks like your code does not encounter this error because your `if` statement is always `False` so that the line looks equivalent to `path_filter = df.Name.str.startswith(path)`.

Therefore, I have these 2 questions:
 - When is `path` supposed to be equal to `TAG_ALL`?
 - Can the line `path_filter = df.Name.str.startswith(path)` be enough?

christophe-david (changes requested):
Thanks for the fix.

It works as expected in my tests, but there are some points to investigate. Please see the details.

src/fastoad/gui/variable_viewer.py
```diff
@@ -249,6 +250,7 @@ def _update_items(self, change=None):
                 if modules_item:
                     # Check if the item exists already
                     if i == len(self._filter_widgets):
+                        # print(i, modules_item)
```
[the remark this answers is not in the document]
ScottDelbecq: Corrected in b6a05d8

src/fastoad/gui/variable_viewer.py
```diff
@@ -228,6 +228,7 @@ def _render_sheet(self) -> display:
         self._filter_widgets = []
         modules_item = sorted(self._find_submodules(self.dataframe))
         if modules_item:
+            modules_item.insert(0, TAG_ALL)
             w = widgets.Dropdown(options=modules_item)
             self._filter_widgets.append(w)
```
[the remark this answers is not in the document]
ScottDelbecq: I agree!
Corrected in 428da5e.

src/fastoad/gui/variable_viewer.py
ScottDelbecq: Thanks indeed I think I got mixed up with other programming languages ...
Yes I guess `path` always contains `TAG_ALL` now.
Your proposition works and seems enough for our usage.
Corrected in fe4c339.

christophe-david: Thanks. Actually, in my tests, `path` never contains `TAG_ALL` 😅 
Yet, `path` can be empty, but it is not a problem here.

christophe-david (approved):

PR #461, opened 2022-10-10
Issue 460 - Optimization viewer fix for 0.0 bound values

### LINKED ISSUE #460: Optimizer viewer save deletes bounds for 0.0
**Describe the bug**
When the save button of the optimizer viewer is used after a modification, the bounds with `0.0` in the optimization section are deleted and converted to `Nan`.

ScottDelbecq:
This PR resolves #460 by changing the manner to that bounds were not `NaN` in optimization viewer to avoid 0.0 to be interpreted as `NaN`.


src/fastoad/gui/optimization_viewer.py
```diff
@@ -199,11 +199,11 @@ def _update_optim_variable(variable: Variable, optim_definition: Dict):
...
-        if meta["lower"] and not isnan(meta["lower"]):
+        if meta["lower"] is not None and not isnan(meta["lower"]):
             optim_definition[section_name][name].update({"lower": meta["lower"]})
         else:
             optim_definition[section_name][name].pop("lower", None)
-        if meta["upper"] and not isnan(meta["upper"]):
+        if meta["upper"] is not None and not isnan(meta["upper"]):
             optim_definition[section_name][name].update({"upper": meta["upper"]})
         else:
             optim_definition[section_name][name].pop("upper", None)
```
christophe-david, line 209: I known, I'm picky, but when I see twice the same fix, I have to ask myself if it cannot be factorized.
What do you think of the suggestion below? (note that GitHub should not be able to apply it as a commit because there is a deleted line in the mix)
```suggestion
        for bound in ["lower", "upper"]:
            if meta[bound] is not None and not isnan(meta[bound]):
                optim_definition[section_name][name].update({bound: meta[bound]})
            else:
                optim_definition[section_name][name].pop(bound, None)
```

ScottDelbecq: I can only agree :)
Corrected in 53169f4

christophe-david (approved):
Thanks for this fix.

PR #430, opened 2022-04-04
Enhancements in Mission module

christophe-david:
Closes #404.

- Documentation has been updated.  Preview [here](https://fast-oad.readthedocs.io/en/mission-fixes/documentation/mission_module/index.html)
  - Description of mission definition and segments has been update
  - A section has been added about extensibility. Description of FlightPoint class has been moved in there.
- `FlightSegment._get_gamma_and_acceleration()` now take a FlightPoint as input, to allow the calculation to depend on more parameters than mass, drag and thrust
- a parameter with a variable as value can now be associated to a unit and a default value that will be used in the OpenMDAO input declaration (and be in generated input data file)
- a target parameter can be declared as relative to the start point of the segment by prefixing the parameter name with `delta_`
- when setting a parameter, a minus sign can be put before a variable name to get the opposite value (can be useful with relative values)
- a parameter can now be set at route or mission level
- dISA can now be set in mission definition file with `isa_offset`
- taxi-out and takeoff are now set in mission definition file:
  - mission starting point (altitude, speed, mass) can now be set using the "start" segment
  - the mass input of the mission can be set using the "mass input" segment. This segment can be anywhere in the mission, though it is expected that fuel consumption in previous segments is mass-independent.
- a mission phase can now contain other phases 
- if a segment parameter (dataclass field) is an array or a list, the associated variable in mission file will be declared with `shape_by_conn=True`

PR #462, opened 2022-10-12
Mission backward compatibility

christophe-david:
This PR completes PR #430 by putting back compatibility with previous mission definitions, that had to assume an automatic addition of taxi-out & pseudo-takeoff phases at the beginning of the mission.

Sphinx doc preview: https://fast-oad.readthedocs.io/en/mission-backward-compatibility/documentation/mission_module/index.html
(This was the occasion to spot a problem with environment creation in RTD. The change in docs/requirements.txt is a temporary fix)


esnguyenvan (approved):
Agree with fix and added documentation

PR #467, opened 2022-11-16
Mission fuel as objective

### LINKED ISSUE #438: Fuel as an objective for the mission builder
As of now, it looks like the only objective we can set for the mission computation is the range of the mission. This makes application like the computation of a payload range diagram or the study, for a constant MTOW, of a change in structural weight  harder to conduct

Instead of having only the range as an attribute of the routes in the mission file, we could have an objective attribute that could be set either as a range or a fuel. In the latter case, the range would be computed so that the fuel consumed during the whole mission matched the objective fuel.

The alternative considered so far is either to use Breguet equation where range and fuel can both be used as objectives but the formula lacks accuracy, or iterating on the range to get the right fuel consumption.

christophe-david:
This PR resolves #438.

It adds the `use_all_block_fuel` boolean parameter to mission definition in mission definition file.

A new `Mission` class has been created to handle that. To avoid confusion, the previous `Mission` class, that was the OpenMDAO group for the job, has been renamed `OMMission`.

The computation of final reserve is now done in this new `Mission`, which was needed for the computation of possible distance with given block fuel. The global behavior is equivalent.

Documentation as been updated. Preview [here](https://fast-oad.readthedocs.io/en/issue-438_mission-fuel-as-objective/documentation/mission_module/mission_file/mission_file.html#mission-definition-section)


docs/documentation/mission_module/mission_file/mission_file.rst
```diff
@@ -203,9 +203,16 @@ Mission definition section
...
+A mission is identified by its name and has 3 attributes:
+
+    - :code:`parts`: list of the :ref:`phase<phase-section>` and/or :ref:`route<route-section>`
+      names that compose the mission, with optionally a last item that is the :code:`reserve`
+      (see below).
+    - :code:`use_all_block_fuel`: if True, the range of the main :ref:`route <route-section>`
+      of the mission will be adjusted so that all block fuel (provided as input
+      `data:mission:<mission_name>:block_fuel`) will be consumed for the mission, excepted the
+      reserve, if defined. The provided range for first route is then used as a first guess
+      to initiate the iterative process.
```
ScottDelbecq, line 215:
```suggestion
      reserve, if defined. The provided range for first route is overridden but used as a first guess
      to initiate the iterative process.
```

src/fastoad/models/performances/mission/tests/test_routes.py
```diff
@@ -203,174 +82,10 @@ def test_ranged_route(low_speed_polar, high_speed_polar, cleanup):
     )
     flight_points = flight_calculator.compute_from(start)
 
-    # plot_flight(flight_points, "test_ranged_flight.png")
+    # plot_flight(flight_points, "test_ranged_flight.png", RESULTS_FOLDER_PATH)
```
ScottDelbecq, line 85: Are you willing to keep this comment for latter debug purposes?

ScottDelbecq (changes requested):
Thanks for this PR! Another big work on the mission analysis...
I have a very small suggestion for the documentation. The rest is very clear for me :)

src/fastoad/models/performances/mission/tests/test_routes.py
christophe-david: Yes.
I grant ourselves some more freedom in test code 😄 .

ScottDelbecq (approved):

PR #485, opened 2023-03-28
New mission registering mechanism

christophe-david:
In preparation for the integration of ground effects into mission definition files, the flight segment recording mechanism has been modified.

It is now based on decorators, while retaining the previous inheritance-based method.

The use of decorators could allow this mechanism to be used for FAST-OAD modules, which could provide some simplification.

Additionally, `FlightSequence` class no longer inherits from `list`.
In the first place, this is because a list-derived class does not mix well with the `deprecated` library. But anyway, it seems safer to have only the desired methods for the `FlightSequence` class instead of having all methods of `list`.


esnguyenvan (approved):
Hello, thanks for this contribution. Compare to the previous mechanism, I find it easier to understand. Having a common mechanism for modules and mission segments registration indeed sounds good (if possible).

PR #497, opened 2023-07-19
Lower tolerance in segment target search

esnguyenvan:
Added test to recreate a bug in the climb segment causing the aircraft to climb indefinitely when it gets just a little higher than MAXIMUM_FLIGHT_LEVEL when trying to reach OPTIMAL_FLIGHT_LEVEL and implemented the fix (lowered tolerance of root_scalar function looking for the moment at which target is reached).


christophe-david (approved):
That will do as a temporary fix.

The core problem is the management of the optimal flight level. There is something cheesy here. I guess the correct way to do that would be to determine the optimal altitude and then use the matching flight level, but it cannot be done so easily.

PR #498, opened 2023-08-24
Enhanced fix for climb segment going far too high when asked for optimal altitude in some cases

christophe-david:
This PR is a simple follow-up to #497.
It applies the same fix, while making it more robust (hopefully).


esnguyenvan (approved):
I agree with the solution, thanks for this improvement

PR #500, opened 2023-08-28
Added documentation for source data file generation

florentLutz:
A documentation on the commands introduced in PR #477 was forgotten and consequently added here. I propose to put the section on source file generation after the section on the generation of input because this is where source file are first mentionned as a way to "fill" an input file. Additionally a reference to a former section was not working, it should be fixed.


christophe-david (changes requested):
Thanks for this addition.

I suggest 2 very minor changes.

florentLutz:
Fixed the ASCII tree and also used one of the less verbose option you proposed botn on the link to the section on how to generate a conf file but also to the section on how to generate a source data file.

christophe-david (approved):
Ok for me. Many thanks.

PR #502, opened 2024-01-12
Setting group options from configuration file

christophe-david:
This PR allows setting values of options for an OpenMDAO group.

In particular, it will allow using the `auto_order` feature introduced by OpenMDAO 3.28.

As for FAST-OAD modules, options can be set simply by a `<key>: <value>` in the corresponding section.

This is a bit inconsistent with the settings of `linear_solver `and `nonlinear_solver`, that are attributes and not options, and will yet be at the same level in the configuration file. But it looked like it was worth privileging the consistency with the setting of module options.

Documentation has been updated and can be previewed [here](https://fast-oad.readthedocs.io/en/group-options/documentation/usage.html#problem-definition).

### DIFF
--- a/docs/documentation/usage.rst
+++ b/docs/documentation/usage.rst
@@ -158,6 +158,10 @@ Problem definition
 
 .. code:: yaml
 
+    model:
+      nonlinear_solver: om.NonlinearBlockGS(maxiter=100, atol=1e-2)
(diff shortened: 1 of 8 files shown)


ScottDelbecq (approved):
Thank you for this PR!
No comments.

PR #503, opened 2024-01-15
Avoiding unnecessary setup operations

christophe-david:
This PR fixes some performance problems of FAST-OAD.

To get knowledge of the structure of OpenMDAO problems, a setup operation is needed. Until now, due to the several changes that were done separately, this setup operation was done almost each time information was needed, which proved to be CPU costly. 

Here the problem analysis is done in only one place, which allows to reduce to the minimum the number of setup calls. There is still one "additional" setup operation needed to analyze the problem, and possibly another one if dynamically shaped variables are inputs of the problem.

Also, `write_needed_inputs` implementation has been put back in `FASTOADPoblem` (while keeping the `FASTOADProblemConfigurator` method), which allows to spare 2 setup calls when called on the problem instance that will do the computation (when done from `FASTOADProblemConfigurator`, it is done on a separate instance).

As a result, the integration test `test_oad_process` uses now 2 setup calls, instead of 10 previously.

### DIFF
--- a/src/fastoad/io/configuration/configuration.py
+++ b/src/fastoad/io/configuration/configuration.py
@@ -21,17 +21,15 @@ from abc import ABC, abstractmethod
 from importlib.resources import open_text
 from typing import Dict
 
-import numpy as np
 import openmdao.api as om
 import tomlkit
 from jsonschema import validate
 from ruamel.yaml import YAML
 
 from fastoad._utils.files import make_parent_dir
-from fastoad.io import DataFile, IVariableIOFormatter
+from fastoad.io import IVariableIOFormatter
 from fastoad.module_management.service_registry import RegisterOpenMDAOSystem, RegisterSubmodel
 from fastoad.openmdao.problem import FASTOADProblem
-from fastoad.openmdao.variables import VariableList
 from . import resources
 from .exceptions import (
     FASTConfigurationBadOpenMDAOInstructionError,
@@ -115,6 +113,10 @@ class FASTOADProblemConfigurator:
 
         problem = FASTOADProblem()
         self._build_model(problem)
+
+        if self._configuration_modifier:
+            self._configuration_modifier.modify(problem)
+
         problem.input_file_path = self.input_file_path
         problem.output_file_path = self.output_file_path
 
@@ -131,9 +133,6 @@ class FASTOADProblemConfigurator:
 
         if read_inputs:
             self._add_design_vars(problem.model, auto_scaling)
-
-        if self._configuration_modifier:
-            self._configuration_modifier.modify(problem)
 
         return problem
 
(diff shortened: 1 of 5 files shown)


src/fastoad/io/configuration/configuration.py
```diff
@@ -115,6 +113,10 @@ def get_problem(self, read_inputs: bool = False, auto_scaling: bool = False) ->
 
         problem = FASTOADProblem()
         self._build_model(problem)
```
florentLutz, line 115: It is not a change in the PR but I'm wondering here if this line means that we rebuild the problem each time we call the `get_problem` method. For instance if we were to keep calling the `write_needed_inputs` from the `FASTOADProblemConfigurator` class we would construct the problem twice.

src/fastoad/openmdao/_utils.py
```diff
@@ -16,52 +16,57 @@
...
 
+T = TypeVar("T", bound=om.Problem)
 
-@contextmanager
-def problem_without_mpi(problem: om.Problem) -> om.Problem:
-    """
-    Context manager that delivers a copy of the given OpenMDAO problem.
 
-    A deepcopy operation may crash if problem.comm is not pickle-able, like a
+def get_problem_copy_without_mpi(problem: T) -> T:
```
florentLutz, line 30: I'm not sure about the name of the function. From the description, I seem to understand that this function allows to avoid the problem with MPI by removing the non-pickable object beforehand and readding it afterwards, which would work regardless of whether the problem has MPI or not. So the `wihout_mpi` seems more like a `mpi_safe` ?

src/fastoad/openmdao/problem.py
```diff
@@ -77,28 +85,56 @@ def setup(self, *args, **kwargs):
...
-                    ivc = om.IndepVarComp()
-                    for name, meta in vars_metadata.items():
-                        # We use a (2,)-shaped array as value here. This way, it will be easier
-                        # to identify dynamic-shaped data in an input file generated from current
-                        # problem.
-                        ivc.add_output(name, [np.nan, np.nan], units=meta["units"])
-                    self.model.add_subsystem(SHAPER_SYSTEM_NAME, ivc, promotes=["*"])
+        self.self_analysis()
+
+        self._analysis.fills_dynamically_shaped_inputs(self)
```
florentLutz, line 90: Isn't this line already ran when you do a `ProblemAnalysis` ?

src/fastoad/openmdao/problem.py
```diff
@@ -176,68 +231,28 @@ def _read_inputs_without_setup_done(self):
...
-            io_metadata = system.get_io_metadata("input")
-            dynamic_vars.update(
-                {
-                    meta["prom_name"]: meta
-                    for name, meta in io_metadata.items()
-                    if meta["shape_by_conn"] and meta["prom_name"] not in output_var_names
-                }
-            )
-        return dynamic_vars
+        self.model.set_order([subsystem_name] + self._analysis.subsystem_order)
```
florentLutz, line 255: Again not linked with this PR, but do we need to reorder the subsystems ? Isn't naturaly done when we add the models  ?

florentLutz (commented):
A lot of comments but nothing that prevents the merging of this PR, just a few questions about how things are done :smile:

src/fastoad/io/configuration/configuration.py
christophe-david: Indeed, the problem is built each time `get_problem` is called. I think it's better like this because:
- if a user called `get_problem`, he is able to keep the generated instance, and I think he has no reason to call `get_problem` again unless he wants a fresh new instance.
- we could imagine a case where the configuration is read, `get_problem` is called, then the configuration is programmatically modified (e.g. optimization definition) and `get_problem` is called again. In such case, a fresh new instance is really better.

This is why `write_needed_inputs` implementation has been moved to the `FASTOADProblem` class, which allows doing this operation on the instance that will be used for the computation.

src/fastoad/openmdao/_utils.py
christophe-david: Good point. Name adopted in 52fe6497df9f02e88896d683a8b1bf954fd1dd21

src/fastoad/openmdao/problem.py
christophe-david: I had hesitated making a property for that, that would control the running of the analysis at the first usage.

Your question made me decide it was the way to go. Modification done in 4eeb488a828b3d679ec208c792b7f1034be563ae.

src/fastoad/openmdao/problem.py
christophe-david: A sure thing is that the component are added in the order of the `add_subsystem` calls.

The fact is that removing this line makes 2 tests fail (these tests do optimization on the Sellar problem).
Other tests (unit and integration) still work fine. 
For the record, with the last OpenMDAO version installed, using `self.model.options["auto_order"] = True` makes the 2 tests work again.

Actually, the 2 tests fail (give incorrect results) because there is no solver at the `Problem.model` level. Indeed, with an IVC in last position, a loop is needed to have the IVC output correctly fill the needed inputs.

florentLutz (approved):
No more comments, thanks for the PR 😃

PR #510, opened 2024-03-11
Centralized way to set options

christophe-david:
This PR adds the possibility to control the `Problem.model_options` attribute from FAST-OAD configuration file.

This [feature](https://openmdao.org/newdocs/versions/latest/features/core_features/options/options.html#setting-options-throughout-a-problem-model-problem-model-options) has been added in OpenMDAO 3.27, which is now the minimum required version for FAST-OAD.

Documentation has been updated. The preview is available [here](https://fast-oad.readthedocs.io/en/model-options/documentation/usage.html#model-options).


docs/documentation/usage.rst
```diff
@@ -133,6 +133,30 @@ optimization problems. It is defined as in Python when assuming the OpenMDAO con
 
 For more details, please see the OpenMDAO documentation on `drivers <http://openmdao.org/twodocs/versions/latest/features/building_blocks/drivers/index.html>`_.
 
+Model options
+==============
+
+OpenMDAO 3.27 introduced a new way to set options for any component in the problem, using the
+:code:`model_options` attribute of the :code:`Problem` object (see OpenMDAO documentation `here <https://openmdao.org/newdocs/versions/latest/features/core_features/options/options.html#setting-options-throughout-a-problem-model-problem-model-options>`_).
+
+This can be controlled from FAST-OAD configuration file, using for instance:
```
florentLutz, line 142: I'd suggest : "This can be controlled from the configuration file, using for instance"

docs/documentation/usage.rst
```diff
@@ -133,6 +133,30 @@ optimization problems. It is defined as in Python when assuming the OpenMDAO con
...
+
+.. code:: yaml
+
+    model_options:
+      "*":
+        propulsion_id: fastoad.wrapper.propulsion.rubber_engine
+        use_xfoil: true
```
florentLutz, line 149: I think the example of the propulsion idea is very pertinent, but I wonder if another example could also be added that does not use the wildcards

docs/documentation/usage.rst
```diff
@@ -133,6 +133,30 @@ optimization problems. It is defined as in Python when assuming the OpenMDAO con
...
+
+With above lines, any OpenMDAO component in the FAST-OAD problem, that has any of these options,
+will use the provided value for the option.
```
florentLutz, line 152: For consistency with an earlier sentence, instead of saying "any OpenMDAO component in the FAST-OAD problem", I'd say: "any components in the problem"

docs/documentation/usage.rst
```diff
@@ -133,6 +133,30 @@ optimization problems. It is defined as in Python when assuming the OpenMDAO con
...
+
+.. note::
+
+  - Please note that the wildcards have to be (double) quoted.
+  - This feature is especially convenient to set options for sub-components of the declared models,
+    since these options are not directly accessible from the configuration file.
+
```
florentLutz, line 159: Still for consistency: I can see that there are other notes on the "Usage" page of the documentation, however, they do not use the same formatting (for instance the note on the XDSM section uses italics), shouldn't we take the opportunity to have the same format everywhere ?

src/fastoad/io/configuration/tests/data/conf_sellar_example/functions.py
```diff
@@ -41,6 +41,11 @@ def setup(self):
 class FunctionF(om.ExplicitComponent):
     """An OpenMDAO component to encapsulate Functions discipline"""
 
+    def initialize(self):
+        # These options have no effect and are used for checks
+        self.options.declare("dummy_f_option", types=int, default=0)
+        self.options.declare("dummy_generic_option", types=str, default="")
+
```
florentLutz, line 48: That may be me being too picky, but if the generic option resembles something like the propulsion_id, should in also be in `FunctionFAlt`

florentLutz (commented):
Thank you for this PR. I was initially unsure whether this functionnality was worth locking a new minimal version of OpenMDAO but after seeing its use for the propulsion id I'm convinced. In addition to my remarks, a small question on the location of the model_options section on the Usage page of the documentation and in the examples. I know it doesn't really matter in terms of how it works, but regarding its "importance", shouldn't it be after the models ?

src/fastoad/io/configuration/tests/data/conf_sellar_example/functions.py
christophe-david: Granted :) (62e98f001f07690301be7b7e1bc097ce85b636d2)

florentLutz (approved):

PR #522, opened 2024-05-03
Enabling pathlib.Path inputs for DataFile objects

### LINKED ISSUE #520: Make DataFile compatible with `pathlib`
As of right now, it is not possible to use a path obtained with `pathlib` to create a DataFile, which makes using a `FASTOADProblem` as well. This comes from the fact that DataFile checks type of the path to be used and only supports `str`, `IO` or `list`. Paths coming from `pathlib`, however have their own specific type

Either add an if to allow `pathlib` paths or transition from `os` to `pathlib` alltogether

Keep using `os` 😅

christophe-david:
Fixes #520 by making the DataFile class able to use `pathlib.Path` instances as input.

Also added some refactoring that CodeClimate was demanding.

### DIFF
--- a/src/fastoad/cmd/api.py
+++ b/src/fastoad/cmd/api.py
@@ -2,7 +2,7 @@
 API
 """
 #  This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
-#  Copyright (C) 2022 ONERA & ISAE-SUPAERO
+#  Copyright (C) 2024 ONERA & ISAE-SUPAERO
 #  FAST is free software: you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
 #  the Free Software Foundation, either version 3 of the License, or
@@ -22,9 +22,10 @@ import sys
 import textwrap as tw
 from collections import defaultdict
 from collections.abc import Iterable
-from time import time
-from typing import Dict, IO, List, Union
 from enum import Enum
+from io import TextIOBase
+from time import time
+from typing import Dict, List, Union
 
 import openmdao.api as om
 import pandas as pd
@@ -328,7 +329,7 @@ def generate_inputs(
 
 def list_variables(
     configuration_file_path: str,
-    out: Union[IO, str] = None,
+    out: Union[TextIOBase, str] = None,
     overwrite: bool = False,
     force_text_output: bool = False,
     tablefmt: str = "grid",
@@ -436,7 +437,7 @@ def _generate_table_format(variables_df, tablefmt="grid"):
 
 def list_modules(
     source_path: Union[List[str], str] = None,
-    out: Union[IO, str] = None,
+    out: Union[TextIOBase, str] = None,
     overwrite: bool = False,
     verbose: bool = False,
     force_text_output: bool = False,
--- a/src/fastoad/cmd/api.py
+++ b/src/fastoad/cmd/api.py
@@ -23,9 +23,9 @@ import textwrap as tw
 from collections import defaultdict
 from collections.abc import Iterable
 from enum import Enum
-from io import TextIOBase
+from os import PathLike
 from time import time
-from typing import Dict, List, Union
+from typing import Dict, List, Union, TextIO
 
 import openmdao.api as om
 import pandas as pd
@@ -328,8 +328,8 @@ def generate_inputs(
 
 
 def list_variables(
-    configuration_file_path: str,
-    out: Union[TextIOBase, str] = None,
+    configuration_file_path: Union[str, PathLike],
+    out: Union[str, PathLike, TextIO] = None,
     overwrite: bool = False,
     force_text_output: bool = False,
     tablefmt: str = "grid",
@@ -436,8 +436,8 @@ def _generate_table_format(variables_df, tablefmt="grid"):
 
 
 def list_modules(
-    source_path: Union[List[str], str] = None,
-    out: Union[TextIOBase, str] = None,
+    source_path: Union[List[Union[str, PathLike]], str, PathLike] = None,
+    out: Union[str, PathLike, TextIO] = None,
     overwrite: bool = False,
     verbose: bool = False,
     force_text_output: bool = False,
--- a/src/fastoad/io/formatter.py
+++ b/src/fastoad/io/formatter.py
@@ -1,5 +1,8 @@
+"""
+Base class for VariableIOFormatter objects.
+"""
 #  This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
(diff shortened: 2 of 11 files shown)


src/fastoad/cmd/api.py
```diff
@@ -328,7 +329,7 @@ def generate_inputs(
 
 def list_variables(
     configuration_file_path: str,
-    out: Union[IO, str] = None,
+    out: Union[TextIOBase, str] = None,
```
florentLutz, line 332: The reason you change from IO to TextIOBase is because the IO type was too generic since we only expect text files ?

src/fastoad/io/formatter.py
```diff
@@ -25,7 +30,7 @@ class IVariableIOFormatter(ABC):
     """
 
     @abstractmethod
-    def read_variables(self, data_source: Union[str, IO]) -> VariableList:
+    def read_variables(self, data_source: Union[str, PathLike, IO, IOBase]) -> VariableList:
```
florentLutz, line 33: Here, why are both `IO` and `IOBase` needed ? Also, why isn't `TextIOBase` used instead of `IO` ?

src/fastoad/io/tests/test_data_file.py
```diff
@@ -76,6 +80,16 @@ def test_datafile_save_read(cleanup, variables_ref):
 
     assert set(data_file_2) == set(variables_ref)
 
+    # Check using text file object --------------------
```
florentLutz, line 83: I see that comments have been put on those part of the test to explain what is done, but there are no comment on the rest of the test. It may be worth adding

src/fastoad/io/variable_io.py
```diff
@@ -11,34 +11,56 @@
...
+from typing import List, Sequence, Union, IO, Optional
 
 from fastoad.openmdao.variables import VariableList
 from . import IVariableIOFormatter
 from .xml import VariableXmlStandardFormatter
+from .._utils.files import as_path
 from ..exceptions import FastError
 
+_LOGGER = logging.getLogger(__name__)  # Logger for this module
+
```
florentLutz, line 28: The _LOGGER object does not seem to be used in the rest of the file, needs to be removed

src/fastoad/io/variable_io.py
```diff
@@ -11,34 +11,56 @@
 #  You should have received a copy of the GNU General Public License
 #  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import logging
```
florentLutz, line 14: Same remark as for the _LOGGER

src/fastoad/io/variable_io.py
```diff
@@ -11,34 +11,56 @@
...
     """
 
-    def __init__(self, data_source: Union[str, IO], formatter: IVariableIOFormatter = None):
+    def __init__(
+        self,
+        data_source: Optional[Union[str, PathLike, IO, IOBase]],
+        formatter: IVariableIOFormatter = None,
+    ):
+        if isinstance(data_source, (str, PathLike)):
+            data_source = as_path(data_source)
```
florentLutz, line 47: Not really a remark, but a question on the definition of the `as_path` method. Why are there 3 definitions with an `@overload` decorator

florentLutz: And now a remark: this method does not change the argument if it's already a path so why include the `PathLike` type in the if ?

src/fastoad/io/variable_io.py
```diff
@@ -11,34 +11,56 @@
...
+        self.formatter = formatter
+
+    @property
+    def formatter(self) -> IVariableIOFormatter:
+        """Class that determines the file format to be used."""
+        return self._formatter
+
+    @formatter.setter
+    def formatter(self, formatter: IVariableIOFormatter):
+        self._formatter = formatter if formatter else VariableXmlStandardFormatter()
```
florentLutz, line 61: Any other reason behind this change other than making it cleaner ?

florentLutz (changes requested):
Some small questions and a required change for the removal of the unused `_LOGGER`, otherwise all good !

src/fastoad/cmd/api.py
christophe-david: The reason is that I noticed that given `f = open(...)`, `isinstance(f, IO)` returns `False` whereas `isinstance(f, IOBase)` returns `True` (and here `TextIOBase` to, indeed, enforce text files).
So it looked like a good idea to change that.

You noticed an inconsistency with other parts of the code, where I kept `IO`. The reason is that I realized later that PyCharm type checker was complaining unless `IO` was in the type hint.

I did some tests with mypy: it never complains when `IO` is in the type hint, but can complain for the type `IOBase` if provided `sys.stdout`

I think now I will get back to having `IO` in type hints. Looks like the expected way. We just have to keep in mind that `isinstance` checks must not use `IO` (which is rather odd, compared to other type hints, but I/O flows have always been a messy subject in Python, I guess 😅).

src/fastoad/io/variable_io.py
christophe-david: > Not really a remark, but a question on the definition of the `as_path` method. Why are there 3 definitions with an `@overload` decorator
It allows to specify different signatures. Here, it allows to have different return types according to the input type.
It is somehow overkill, but I wanted to test this feature :)

christophe-david: > And now a remark: this method does not change the argument if it's already a path so why include the `PathLike` type in the if ?
Strictly speaking, a PathLike object can be something else than a Path object. It is just guaranteed that a Path object can be built from a PathLike object.

On a purely pragmatic point of view, to avoid the `if` statement, `as_path()` could return the original object if it cannot be converted to a Path object. Currently, in such case, it returns None.
The current way seems a more intuitive behavior to me, given the name of the function, so I'm still undecided...

src/fastoad/io/variable_io.py
christophe-david: Yes. It allows to get back to the default formatter when setting the property to None.
The DataFile class has been reworked and can now set the formatter after having instantiated its internal VariableIO object.

src/fastoad/io/variable_io.py
christophe-david: Fixed in a4b8a4bf354349003ee297fc5a891d1c7464f0f3

src/fastoad/io/variable_io.py
christophe-david: Fixed in a4b8a4bf354349003ee297fc5a891d1c7464f0f3

PR #526, opened 2024-06-07
Climb was not stopping when start was already over the asked optimal altitude/flight level

christophe-david:
In mission, when asking a climb up to an optimal altitude/flight level (i.e. where the maximum CL is achieved), if the maximum CL was achieved for a lower altitude than the starting one, the climb was not stopped until hard limits were attained (the 40000m hard-coded limit, or possibly a crash of the propulsion model).

This PR fixes this problem.


esnguyenvan (approved):
Thanks for this PR. Tested locally on a similar problem, the fix fixes well.

PR #530, opened 2024-06-13
Fix behavior when model_options could add input variables

christophe-david:
This PR fixes a bug that was occurring when an option in model_options could add input variables to some component : the generation of the input file was Ok, but the added input variables were not read for problem evaluation.


esnguyenvan (approved):
Thanks for the fix and the added test. It seems all good

PR #532, opened 2024-06-18
Better NaN detection in inputs

christophe-david:
FAST-OAD behavior was to raise immediately an error if the input file contains a Nan value (for an actually used input variable).

This PR modifies the behavior so an error is raised also when an input variable with a NaN default value is not present in input file.

### DIFF
--- a/src/fastoad/openmdao/exceptions.py
+++ b/src/fastoad/openmdao/exceptions.py
@@ -21,20 +21,6 @@ from fastoad._utils.files import as_path
 from fastoad.exceptions import FastError
 
 
-#  This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
-#  Copyright (C) 2024 ONERA & ISAE-SUPAERO
-#  FAST is free software: you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation, either version 3 of the License, or
-#  (at your option) any later version.
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#  You should have received a copy of the GNU General Public License
-#  along with this program.  If not, see <https://www.gnu.org/licenses/>.
-
-
 class FASTOpenMDAONanInInputsError(FastError):
     """Raised if NaN values are read in input data file."""
 
--- a/src/fastoad/openmdao/exceptions.py
+++ b/src/fastoad/openmdao/exceptions.py
@@ -1,10 +1,28 @@
 """
 Module for custom Exception classes linked to OpenMDAO
 """
+#  This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
+#  Copyright (C) 2024 ONERA & ISAE-SUPAERO
+#  FAST is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from os import PathLike
+from typing import Iterable
+
+from fastoad._utils.files import as_path
+from fastoad.exceptions import FastError
 
 
 #  This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
-#  Copyright (C) 2022 ONERA & ISAE-SUPAERO
+#  Copyright (C) 2024 ONERA & ISAE-SUPAERO
 #  FAST is free software: you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
 #  the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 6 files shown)


src/fastoad/openmdao/exceptions.py
```diff
@@ -16,22 +14,37 @@
...
+#  FAST is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
florentLutz, line 35: The header/licence has been duplicated in the same script, it can be removed.

src/fastoad/openmdao/exceptions.py
```diff
@@ -16,22 +14,37 @@
...
+
+
+class FASTOpenMDAONanInInputsError(FastError):
```
florentLutz, line 38: I know it was already there in the previous exception name but why does it start with FASTOpenMDAO ? Is it FAST or OpenMDAO. Also being very picky about things that don't matter, since NAN is already an acronym (like MDAO) why isn't it all uppercase ?

src/fastoad/openmdao/exceptions.py
```diff
@@ -16,22 +14,37 @@
...
     """Raised if NaN values are read in input data file."""
 
-    def __init__(self, input_file_path: str, nan_variable_names: List[str]):
-        self.input_file_path = input_file_path
-        self.nan_variable_names = nan_variable_names
+    def __init__(self, input_file_path: [str, PathLike], nan_variable_names: Iterable[str]):
+        self.input_file_path = as_path(input_file_path)
+        self.nan_variable_names = sorted(list(nan_variable_names))
```
florentLutz, line 43: Why does the list need to be sorted ?

src/fastoad/openmdao/exceptions.py
```diff
@@ -16,22 +14,37 @@
...
 
-        msg = "NaN values found in input file (%s). Please check following variables: %s" % (
-            input_file_path,
-            nan_variable_names,
+        msg = (
+            f"NaN values found in inputs. Please check that {input_file_path} contains "
+            f"following variables and that they are not NaN: {self.nan_variable_names}"
```
florentLutz, line 47: Not a remark but a question, what is the difference between the two ways of formatting the error message ?

src/fastoad/openmdao/problem.py
```diff
@@ -198,21 +198,44 @@ def _get_problem_inputs(self) -> Tuple[VariableList, VariableList]:
...
+        if non_filled_variable_names:
+            raise FASTOpenMDAONanInInputsError(self.input_file_path, non_filled_variable_names)
 
-        nan_variable_names = [var.name for var in input_variables if np.any(np.isnan(var.value))]
-        if nan_variable_names:
-            raise FASTOpenMDAONanInInputFile(self.input_file_path, nan_variable_names)
+        return input_file_variables, unused_variables
 
-        return input_variables, unused_variables
+    def _get_remaining_nan_variable_names(self, input_file_variables):
```
florentLutz, line 219: Based on the docstring and the way `non_filled_variable_names` is built, it seems like this function return a list of problem inputs whose default is nan and who do not appear in the input of that are set at nan in the input file. in that case, why does the function name contains `reamining` ? It seems like it get all NaN variables

src/fastoad/openmdao/tests/data/missing_inputs.xml
```diff
@@ -0,0 +1,18 @@
...
+  ~ This program is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <https://www.gnu.org/licenses/>.
+  -->
+
+<aircraft>
+  <useless>1.0</useless>
```
florentLutz, line 17: useless or unused ❔

src/fastoad/openmdao/tests/test_problem.py
```diff
@@ -148,19 +150,53 @@ def test_problem_read_inputs_with_nan_inputs(cleanup):
...
+        problem.read_inputs()
+    assert exc_info_1.value.input_file_path == input_data_path
+    assert exc_info_1.value.nan_variable_names == ["z"]
+
+    problem.setup()
+
+    with pytest.raises(FASTOpenMDAONanInInputsError) as exc_info_2:
+        problem.read_inputs()
+    assert exc_info_2.value.input_file_path == input_data_path
+    assert exc_info_2.value.nan_variable_names == ["z"]
```
florentLutz, line 197: Is the test before AND after the setup really needed ?

florentLutz (changes requested):
Thank you for the PR !
Some minor remarks and questions. The only change i firmly requests is the doubled licence in exceptions.py. The rests I would simply like to discuss

src/fastoad/openmdao/exceptions.py
christophe-david: Bad joke that PyCharm does from time to time :/ Fixed in 781349be84f21c6ced8c6769f2a251e49e8f49b3

src/fastoad/openmdao/exceptions.py
christophe-david: Indeed, I didn't pay attention to this part of the name, but it does not make much sense here. Fixed in 72651bf030753de7f83546b2324f70a522186ad1
And for the "picky" part ;), the standard acronym is "NaN", but I feel that using it would make the CamelCase messy.

src/fastoad/openmdao/exceptions.py
christophe-david: Just to ensure the error message will give them in alphabetical order, avoiding the possibility of a messy output message.

src/fastoad/openmdao/exceptions.py
christophe-david: The previously used one if out of fashion ;)
More seriously, this one is now the standard, and I'd want to remove the old one from FAST-OAD (except in `logging` messages), so I dot it when I can.
And IMHO, it is generally  clearer.

src/fastoad/openmdao/problem.py
christophe-david: The idea is getting the variables that "remain" at NaN after reading the input file (which match your description, I guess).
But I would gladfully consider another naming...

src/fastoad/openmdao/tests/data/missing_inputs.xml
christophe-david: So picky 🤣 
But you are right. Fixed in 927f3c8b5004d3d8378e7511ab361bde9a9d02eb

src/fastoad/openmdao/tests/test_problem.py
christophe-david: Given that reading the entries does not involve exactly the same steps in these two cases, I feel more comfortable this way.

src/fastoad/openmdao/problem.py
florentLutz: I do understand the choice of the word remaining now. I don't have any name that are more explicit in mind, maybe simply `_get_nan_variable_names` ?

florentLutz (approved):
Thanks for the changes !

PR #535, opened 2024-06-21
Fixed issue 534

### LINKED ISSUE #534: Change variables displayed by default in the `MissionViewer`
By default when the MissionViewer is displayed, it displays the altitude as a function of the delta to the ISA. It doesn't make a lot of sense in terms of aircraft design. In the comment of the code, however it tells that it displays the altitude as a function of the ground distance which is selected with hard coded index. This tells me that it may have originally done the job but now is not working as intended anymore and that the problem will repeat with each subsequent addition of column in the mission .CSV.

Instead of using hard coded index, look for the right index among the column name

florentLutz:
Made it so that the altitude and ground distance are taken by default when displaying the mission viewer if it is at all possible. Also added corresponding tests to check that it indeed works. This PR closes #534


Mokyoslurp:
The test flight_points.csv files may be a little old since some of the keys of the flight_points.csv generated with newer FAST-OAD versions come with the unit ("altitude [m]", "mass [kg]", etc), so the mission viewer still plots altitude against isa_offset (using default keys 1 and 2) with new flight_points.csv.

florentLutz:
Thank you for being vigilant @Mokyoslurp. Updated the column name we seek to match the format used in the FlightPoint class which is used as a base to write the output csv. Also added some freshly generated .csv file in the test.

src/fastoad/gui/mission_viewer.py
```diff
@@ -73,11 +75,22 @@ def display(self):
...
+        flight_point_units = FlightPoint.get_units()
+
         # By default ground distance
-        self._x_widget = widgets.Dropdown(value=keys[2], options=keys)
+        unit_ground_distance = flight_point_units["ground_distance"]
+        column_ground_distance = f"ground_distance [{unit_ground_distance}]"
+        idx_ground_distance = (
+            keys.to_list().index(column_ground_distance) if column_ground_distance in keys else 3
+        )
+        self._x_widget = widgets.Dropdown(value=keys[idx_ground_distance], options=keys)
```
christophe-david, line 86: I would suggest this more pythonic way:
```suggestion
        x_label = column_ground_distance if column_ground_distance in keys else keys[3]
        self._x_widget = widgets.Dropdown(value=x_label, options=keys)
```
Besides, I would be tempted to add the method `_get_label(self, keys, quantity_name, default_idx)` to replace lines 81-85 by `x_label = self._get_label(keys, "ground_distance", 3)` and lines 90-92 by `y_label = self._get_label(keys, "altitude", 1)`.
That would require to include line 78 in the method, and then do it twice, but I think it's acceptable.

christophe-david (changes requested):
Done by the book, with proper testing. Great.

I just suggest to do the change in a more pythonic way. And possibly a refactoring that would avoid repetition of some code and some string values.

src/fastoad/gui/mission_viewer.py
florentLutz: Done in 7e1f5284

PR #539, opened 2024-06-27
Ruff usage

christophe-david:
Ruff is now used instead of Black and Flake8.

[Wiki](https://github.com/fast-aircraft-design/FAST-OAD/wiki) has been updated.


src/fastoad/io/xml/exceptions.py
```diff
@@ -1,4 +1,4 @@
-""" Exceptions for io.xml module """
+"""Exceptions for io.xml module"""
```
florentLutz, line 1: Small comment here, but there is something I find odd that ruff didn't pick up. Here, this comment doesn't end with a dot while in the __init__ of the gui folder, it does and yet both have passed. Does that mean that there is no convention for that ?

src/fastoad/model_base/tests/test_atmosphere.py
```diff
@@ -155,12 +155,12 @@ def test_speed_conversions():
...
-        [200.0, 197.3624, 116.1973],
-        [270, 161.8971, 266.6984],
-        [400, 395.5782, 252.6355],
-    ]
+    # expected_CAS = [  # currently unused
+    #     [100.0, 98.5799, 56.3313],
+    #     [200.0, 197.3624, 116.1973],
+    #     [270, 161.8971, 266.6984],
+    #     [400, 395.5782, 252.6355],
+    # ]
```
florentLutz, line 163: Since it is unused and since the Atmosphere class will soon be deprecated in the core, can't we just delete it ?

src/fastoad/notebooks/02_From_Python_to_FAST-OAD/modules/OpenMDAO/mass/sub_components/compute_owe_exercise.py
```diff
@@ -1,5 +1,3 @@
-import numpy as np
-
```
florentLutz, line 2: I'm not sure about that deletion since the users of the tutorial might have to use numpy in the formula or the declaration of inputs/outputs

florentLutz (commented):
In the end, I have no issue with the deletion of the upper_value_results but there's still one or two things I'm curious about 😄

src/fastoad/io/xml/exceptions.py
christophe-david: AFAIK, there is indeed no convention about that.
I won't pretend I am consistent on this topic, but I try to end actual sentences by full stop. For "non-sentences comments", I don't feel it is mandatory.

src/fastoad/model_base/tests/test_atmosphere.py
christophe-david: I don't mind. As you say, we should remove the class and the associated tests when we'll decide to rely only on stdatm, so I don't consider that as important.

src/fastoad/notebooks/02_From_Python_to_FAST-OAD/modules/OpenMDAO/mass/sub_components/compute_owe_exercise.py
christophe-david: Fixed in 1116405cee9cb602b3e580b697879a946a7aa985

florentLutz (approved):

Issue #544, opened 2024-07-02
Impossible to hide one aircraft on the first figure of mass_breakdown_bar_plot


Mokyoslurp:
**Describe the bug**
When using the plotly functionnality to hide a set of data on the bar mass breakdown, only the the second figure hides the data. The first one keeps all the data plotted. 

**To Reproduce**
Plot the bar mass breakdown with at least 2 aircraft, then click on the legend on one of the aircraft.

**Expected behavior**
Both figures should hide the selected the aircraft.

**Error message**
Before clicking:
![image](https://github.com/fast-aircraft-design/FAST-OAD/assets/46455364/61eeaa93-ea98-47df-aa08-e07416f9b56e)
After clicking:
![image](https://github.com/fast-aircraft-design/FAST-OAD/assets/46455364/7b7ed8a8-7efb-45bd-8e3f-e2b6a1949962)

**Environment**
- Operating System: All
- Python environment: Anaconda python 3.9

christophe-david:
Thanks for the report.
Indeed, we ask Plotly to show the legend only for the right-handed plot, and we miss using the `legendgroup` attribute. We will fix that.

christophe-david:
My bad, didn't see @aeomath had already proposed a fix 😅

PR #545, opened 2024-07-02
Adding grouped legend for each aircraft configuration (fix issue #544)

aeomath:
This PR should close issue #544 

When using subplots in the mass_breakdown_bar_plot , it was impossible to isolate one aircraft from another and only one plot was masked.

To resolve this ,  we create a grouped legend for each aircraft configuration ( same methods as the one to have the same color for each aircraft)

### DIFF
--- a/src/fastoad/gui/analysis_and_plots.py
+++ b/src/fastoad/gui/analysis_and_plots.py
@@ -316,11 +316,11 @@ def mass_breakdown_bar_plot(
     # Same legend for each aircraft configuration
     l = int(len(fig.data)/2)
     aircraft_config = f"aircraft{l}"
-    
+
     weight_labels = ["MTOW", "OWE", "Fuel - Mission", "Payload"]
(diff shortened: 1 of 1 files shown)


christophe-david (approved):
Many thanks for this contribution!

If it happens you ever open another PR, I would suggest that you use the auto-format tool (currently Black), ideally in association with pre-commit. In case of need, our Wiki provides information about that [here](https://github.com/fast-aircraft-design/FAST-OAD/wiki/Development-environment).

christophe-david:
And be sure having auto-format activated in your IDE will make your life easier. To be noted: we currently use Black, but we will soon switch to Ruff, that has the advantage of doing formatting and linting, and of doing them in a flash.

Also, this PR made me aware that only a part of our test workflows are triggered by external pull requests. And among the missing tests, there is the check of formatting.
Then it is a valuable experience for us also :)

PR #546, opened 2024-07-02
Plots now handle horizontal_tail:center new naming

ScottDelbecq:
This is a minor PR to enable to handle a new geometry naming for horizontal tail (`data:geometry:horizontal_tail:root:chord` &#8594; `data:geometry:horizontal_tail:center:chord`) in the plots.
The modification does not affect the compatibility with old naming

### DIFF
--- a/src/fastoad/gui/analysis_and_plots.py
+++ b/src/fastoad/gui/analysis_and_plots.py
@@ -148,6 +148,7 @@ def aircraft_geometry_plot(
     )
 
     # Horizontal Tail parameters
+    # Keeping backward compatibility for horizontal_tail:root
     if "data:geometry:horizontal_tail:center:chord" not in variables.names():
         ht_root_chord = variables["data:geometry:horizontal_tail:root:chord"].value[0]
     else:
--- a/src/fastoad/gui/analysis_and_plots.py
+++ b/src/fastoad/gui/analysis_and_plots.py
@@ -148,7 +148,10 @@ def aircraft_geometry_plot(
     )
 
     # Horizontal Tail parameters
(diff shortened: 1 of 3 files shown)


src/fastoad/gui/analysis_and_plots.py
```diff
@@ -148,7 +148,10 @@ def aircraft_geometry_plot(
     )
 
     # Horizontal Tail parameters
-    ht_root_chord = variables["data:geometry:horizontal_tail:root:chord"].value[0]
+    if "data:geometry:horizontal_tail:center:chord" not in variables.names():
```
christophe-david, line 151: I think a comment to state we keep backward compatibility here would be good.

And as a side note, `if x in seq:` is generally preferred than `if x not in seq:`.
I mean, your choice here is driven only by code history. If you had coded that from scratch, no doubt you would have done it the other way.

christophe-david (changes requested):
Thanks for the change.

I think a dedicated test would be good, though. Don't hesitate to have a second XML data file and use a loop to test both of them.

Once the next version of CS25 is released, we will have to make use of the already computed global positions, and the second data file will then have several differences with the first one.

src/fastoad/gui/analysis_and_plots.py
ScottDelbecq: Corrected in [54240a5](https://github.com/fast-aircraft-design/FAST-OAD/pull/546/commits/54240a58b538344a127a200d68c3fe13f00df9f0)

christophe-david (approved):
Nicely done.

It's a blind spot on my part that I don't think enough about the parameterize fixture. I should definitely make sure I have that option in mind in the future.

PR #548, opened 2024-07-05
Pre commit testing

enricostragiotti:
This PR solves the problems of using VS code with the official python extension and pre-commit (more info [here](https://github.com/microsoft/vscode-python/issues/10165)). This modification does not impact the functioning of the pre-commit routine as poetry is already needed and used for `sync_doc_requirements`.


christophe-david:
This PR is no longer applicable as the pre-commit configuration has been completely changed to use Ruff.

PR #549, opened 2024-07-08
Mission viewer update_layout

Mokyoslurp:
This PR adds a method to update the mission viewer layout before displaying it.


Mokyoslurp:
You are right, this could be done directly in display method. The way I did it  is a copy of plotly's workflow (update then display), so that I could factorize some code with other plots. The method signature is a pure copy of plotly's (they have that dict1) but I agree this is a bad naming.
Tests are buggy on my VScode for some reason so I wasn't able to run all of them automatically

PR #551, opened 2024-07-12
Using output values in input file

christophe-david:
This PR modifies the input reading process so that values of output variables, if present in input file, are also set before computation.
In some cases, it might speed up the convergence (Sellar problem is solved in 2 iterations).

### DIFF
--- a/src/fastoad/openmdao/problem.py
+++ b/src/fastoad/openmdao/problem.py
@@ -199,12 +199,14 @@ class FASTOADProblem(om.Problem):
         :return: VariableList of needed input variables, VariableList with unused variables.
         """
 
-        problem_inputs_names = [var.name for var in self.analysis.problem_variables if var.is_input]
-
         input_file_variables = DataFile(self.input_file_path)
 
         unused_variables = VariableList(
-            [var for var in input_file_variables if var.name not in problem_inputs_names]
+            [
+                var
+                for var in input_file_variables
+                if var.name not in self.analysis.problem_variables.names()
+            ]
         )
         for name in unused_variables.names():
             del input_file_variables[name]
--- a/src/fastoad/openmdao/tests/test_problem.py
+++ b/src/fastoad/openmdao/tests/test_problem.py
@@ -39,7 +39,7 @@ def cleanup():
     RESULTS_FOLDER_PATH.mkdir(parents=True)
 
 
-def test_write_outputs():
+def test_write_outputs(cleanup):
     problem = FASTOADProblem()
     problem.model.add_subsystem("sellar", SellarModel(), promotes=["*"])
     problem.output_file_path = RESULTS_FOLDER_PATH / "output.xml"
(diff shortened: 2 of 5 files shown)


src/fastoad/openmdao/problem.py
```diff
@@ -199,12 +199,14 @@ def _get_problem_inputs(self) -> Tuple[VariableList, VariableList]:
...
-
         input_file_variables = DataFile(self.input_file_path)
 
         unused_variables = VariableList(
-            [var for var in input_file_variables if var.name not in problem_inputs_names]
+            [
+                var
+                for var in input_file_variables
+                if var.name not in self.analysis.problem_variables.names()
+            ]
```
florentLutz, line 209: If I understand correctly, this is the heart of this PR ? The ouputs of the problem are not considered unused anymore so they will be set in the `_set_input_values_post_setup` or `_set_input_values_pre_setup` methods ? Which raises the question of whether those name should be kept because output values can also be set now. Or maybe those method should be rename to `_set_input_file_values...`

src/fastoad/openmdao/tests/test_problem.py
```diff
@@ -115,6 +116,29 @@ def test_problem_read_inputs_before_setup(cleanup):
     assert_allclose(problem.get_val(name="z", units="m**2"), [4.0, 3.0])
     assert_allclose(problem["f"], 21.7572, atol=1.0e-4)
 
+    # Set values of inputs and outputs
+    # (FASTOADProblem has to defer output setting after setup to have it effective,
+    #  so this test works also for read after setup)
```
florentLutz, line 121: Not to be too picky (at this pont I have to say it every PR) but when you say that FASTOADProbelm has to defer output setting after setup it is more that the only values that are setup before setup are values with shape_by_conn, no ?

florentLutz (commented):
Thank you for the PR ! One smal question and one remark 😄

src/fastoad/openmdao/problem.py
christophe-david: Well, we still talk about input **values**. It's just that they now concern also output **variables** 😉

src/fastoad/openmdao/tests/test_problem.py
christophe-david: This is right. The point was to explain why I don't try to test also a `read_inputs()` operation after `setup()`.
I guess I could make it clearer...

christophe-david: Actually, I realize the point of this comment was a bit different. That proves very well it was not clear enough. Modified in 4f93f70ea6f39a2d2224e2c10c2868151694b86c.
You'll tell me if it's clearer :)
BTW, I quickly modified a test component to have a `shape_by_conn=True` on an output. Maybe not the best test, on a one-element array, but it's a corner case, so it's better than nothing.

florentLutz: Thank you for the added test.
As for the comment, I'm still 100% sure the second part is clear enough may I suggest:
_A FASTOADProblem will automatically set output values after setup using set_val(), which is functionally the same as doing the setup()->read_inputs() sequence._
May I also suggest that, between the problem setup and run_model on line 131 and 132 we assert that the value of variable such as g1 and g2 are not to their default ? Just to showcase what happens behind the scenes ?

florentLutz (commented):
I answered your comments, other than that the PR is approved 😄

florentLutz (approved):
Perfect, thank you for the added explanations

PR #556, opened 2024-08-27
Update to iPOPO 3.x

christophe-david:
Update to take into account version 3.x of iPOPO.

A minor adaption was needed (I don't think it is an intended change in iPOPO, but it's easy to handle anyway)


src/fastoad/module_management/service_registry.py
```diff
@@ -150,7 +150,10 @@ def get_provider_description(cls, instance_or_id: Union[str, T]) -> str:
         :return: the description associated to given instance or identifier
         """
 
-        return cls._get_provider_property(instance_or_id, DESCRIPTION_PROPERTY_NAME)
+        description = cls._get_provider_property(instance_or_id, DESCRIPTION_PROPERTY_NAME)
+        if description is None or isinstance(description, property):
+            description = ""
+        return description
```
florentLutz, line 156: Is there a reason behind moving the check on wether description was None ?

christophe-david: At first, it was just for sake of clarity (one "if" in one place). Also, it's often better to have a "clean" result as soon as possible in the call hierarchy. And it was more consistent with the return type of the method ( "str" and not "Optional[str]").
And thanks to you, I realized I missed to take this change into account on the only other call of this method. I did it in 92e7503f4ca2c10aaed85617c294a4b9c184c7fa.

florentLutz (approved):
All good. Thanks!

PR #558, opened 2024-08-30
Update of Flightpoint class

christophe-david:
This PR adds new features to the FlightPoint class.

Each field of the FlightPoint class is now described as cumulative or not, and present in output or not. The "field descriptor" mechanism is also used to store unit information.

Added fields behave the same.

FlightPoint documentation has been slightly modified, and can be previewed [here](https://fast-oad.readthedocs.io/en/flightpoint-update/documentation/mission_module/extensibility/flight_point.html).

### DIFF
--- a/docs/documentation/mission_module/extensibility/flight_point.rst
+++ b/docs/documentation/mission_module/extensibility/flight_point.rst
@@ -23,12 +23,11 @@ Available flight parameters
 ***************************
 The documentation of :class:`~fastoad.model_base.flight_point.FlightPoint` provides
 the list of available flight parameters, available as attributes.
-As FlightPoint is a dataclass, this list is available through Python using::
+This list is available through Python using::
 
     >>> import fastoad.api as oad
-    >>> from dataclasses import fields
 
-    >>> [f.name for f in fields(oad.FlightPoint)]
+    >>> oad.FlightPoint.get_field_names()
 
 *******************************
 Exchanges with pandas DataFrame
@@ -64,15 +63,20 @@ smoothly, especially when exchanging data with pandas, you have to work at class
 This can be done using :meth:`~fastoad.model_base.flight_point.FlightPoint.add_field`, preferably
 outside of any class or function::
 
-    # Adds a float field with None as default value
-    >>> FlightPoint.add_field("ion_drive_power")
+    # Adding a float field with None as default value
+    >>> FlightPoint.add_field(
+    ...    "ion_drive_power",
+    ...    unit="YW",
+    ...    is_cumulative=False, # Tells if quantity sums up during mission
+    ...    is_output=True, # Tells if quantity is expected as mission output
+    ...    )
 
-    # Adds a field and define its type and default value
+    # Adding a field and defining its type and default value
     >>> FlightPoint.add_field("warp", annotation_type=int, default_value=9)
 
     # Now these fields can be used at instantiation
     >>> fp = FlightPoint(ion_drive_power=110.0, warp=12)
 
-    # Removes a field, even an original one (useful only to avoid having it in outputs)
+    # Removing a field, even an original one
     >>> FlightPoint.remove_field("sfc")
 
--- a/src/fastoad/model_base/flight_point.py
+++ b/src/fastoad/model_base/flight_point.py
@@ -225,6 +225,16 @@ class FlightPoint:
         new_point.scalarize()
         return new_point
 
+    def scalarize(self):
+        """
+        Convenience method for converting to scalars all fields that have a
+        one-item array-like value.
+        """
+        self_as_dict = asdict(self)
+        for field_name, value in self_as_dict.items():
+            if np.size(value) == 1:
+                setattr(self, field_name, np.asarray(value).item())
+
     @classmethod
     def get_field_names(cls):
         """
@@ -319,16 +329,6 @@ class FlightPoint:
             delattr(cls, name)
             del cls.__annotations__[name]
             dataclass(cls)
-
-    def scalarize(self):
-        """
-        Convenience method for converting to scalars all fields that have a
-        one-item array-like value.
-        """
-        self_as_dict = asdict(self)
-        for field_name, value in self_as_dict.items():
-            if np.size(value) == 1:
-                setattr(self, field_name, np.asarray(value).item())
 
     @classmethod
     def _redeclare_fields(cls):
--- a/src/fastoad/model_base/flight_point.py
+++ b/src/fastoad/model_base/flight_point.py
@@ -16,18 +16,18 @@
 from copy import deepcopy
 from dataclasses import asdict, dataclass, field, fields
 from numbers import Number
-from typing import Any, List, Mapping, Optional, Sequence, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
 
 import numpy as np
 import pandas as pd
 
 from fastoad.constants import EngineSetting
 
-FIELD_STATUS = "field_status"
+FIELD_DESCRIPTOR = "field_descriptor"
 
 
 @dataclass
-class _FieldStatus:
+class _FieldDescriptor:
     """
     Class to be used as dataclass field metadata.
     """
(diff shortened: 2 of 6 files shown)


src/fastoad/model_base/flight_point.py
```diff
@@ -12,16 +12,30 @@
...
+
+@dataclass
+class _FieldDescriptor:
+    """
+    Class to be used as dataclass field metadata.
+    """
+
+    is_cumulative: Optional[bool] = False
+    is_output: Optional[bool] = True
+    unit: Optional[str] = None
```
florentLutz, line 37: Don't value and default value also need to be in the metadata ?

docs/documentation/mission_module/extensibility/flight_point.rst
```diff
@@ -64,15 +63,20 @@ smoothly, especially when exchanging data with pandas, you have to work at class
 This can be done using :meth:`~fastoad.model_base.flight_point.FlightPoint.add_field`, preferably
 outside of any class or function::
```
florentLutz, line 64: I suggest removing the "of" on that line to be consistent with the description of the FlightPoint dataclass

src/fastoad/model_base/flight_point.py
```diff
@@ -59,78 +73,127 @@ class FlightPoint:
...
-    mass: float = None  #: Mass in kg.
-    consumed_fuel: float = 0.0  #: Consumed fuel since mission start, in kg.
-    true_airspeed: float = None  #: True airspeed (TAS) in m/s.
-    equivalent_airspeed: float = None  #: Equivalent airspeed (EAS) in m/s.
-    mach: float = None  #: Mach number.
-    engine_setting: EngineSetting = None  #: Engine setting.
+    #: Time in seconds.
+    time: float = field(
+        default=0.0, metadata={FIELD_DESCRIPTOR: _FieldDescriptor(is_cumulative=True, unit="s")}
+    )
```
florentLutz, line 104: Is this added complexity necessary or is it just cleaner to do it this way ?

src/fastoad/model_base/flight_point.py
```diff
@@ -59,78 +73,127 @@ class FlightPoint:
...
+
+    #: slope angle derivative in rad/s
+    slope_angle_derivative: float = field(
+        default=None, metadata={FIELD_DESCRIPTOR: _FieldDescriptor(unit="rad/s")}
     )
 
+    #: Name of current phase.
+    name: str = field(default=None, metadata={FIELD_DESCRIPTOR: _FieldDescriptor()})
+
+    __field_descriptors = {}  # Will store field metadata when needed. Must be accessed through
```
florentLutz, line 195: Must be accessed throught what ?

src/fastoad/model_base/flight_point.py
```diff
@@ -175,21 +238,31 @@ def make_absolute(self, reference_point: "FlightPoint") -> "FlightPoint":
         :return: the copied flight point with no relative field.
         """
         new_point = deepcopy(self)
```
florentLutz, line 240: What does absolute and relative means for fields ?

src/fastoad/model_base/flight_point.py
```diff
@@ -198,7 +271,45 @@ def get_units(cls) -> dict:
...
+        }
+
+    @classmethod
+    def get_unit(cls, field_name) -> Optional[str]:
+        """
+        Returns unit for asked field.
+
+        A dimensionless physical quantity will have "-" as unit.
+        """
+        return cls._get_field_descriptors().get(field_name, _FieldDescriptor()).unit
```
florentLutz, line 286: If I understand correctly, this syntax makes it so that if the field_name is not in the __field_descriptors we get a None ? Wasn't there a more readable way ?

florentLutz: Also why not _FieldDescriptor(None, None, None) as is done below ?

src/fastoad/model_base/flight_point.py
```diff
@@ -232,17 +351,28 @@ def add_field(cls, name: str, annotation_type=float, default_value: Any = None,
         :param name: field name
         :param annotation_type: field type
         :param default_value: field default value
-        :param unit: expected unit for the added field ("-" should be provided for a dimensionless
-                     physical quantity)
+        :param unit: expected unit for the added field. "-" should be provided for a dimensionless
+                     physical quantity. Set to None, when unit concept does not apply.
+        :param cumulative: True if field value is sums up during mission
```
florentLutz, line 356: Is summed up ?

src/fastoad/model_base/flight_point.py
```diff
@@ -252,19 +382,38 @@ def remove_field(cls, name):
...
             del cls.__init__  # Delete constructor to allow it being rebuilt with dataclass() call
             delattr(cls, name)
             del cls.__annotations__[name]
             dataclass(cls)
-            if name in cls._units:
-                del cls._units[name]
 
-    def scalarize(self):
+    @classmethod
+    def _get_field_descriptors(cls) -> Dict[str, _FieldDescriptor]:
```
florentLutz, line 392: Can't this be done with getters and setters ?

src/fastoad/model_base/flight_point.py
```diff
@@ -252,19 +382,38 @@ def remove_field(cls, name):
...
+
+        In current state, fields are now declared using only default value (no metadata). We have
+        to redo the field declaration properly.
+        """
+        for cls_field in fields(cls):
+            setattr(
+                cls, cls_field.name, field(default=cls_field.default, metadata=cls_field.metadata)
+            )
+
+        cls.__field_descriptors = {}  # Will need to rebuild this dict on next usage.
```
florentLutz, line 419: I don't really understand why this method (and more generally, the deletion of the constructor and other attributes) is necessary. would you mind explaining it to me ?

src/fastoad/model_base/tests/test_flight_point.py
```diff
@@ -80,3 +80,37 @@ def test_scalarize():
     assert fp.mass == 70000.0
     assert fp.altitude == 1000.0
     assert_allclose(fp.mach, [0.7, 0.8])
+
+
+def test_descriptors():
+    FlightPoint.add_field(
+        "foo", annotation_type=float, default_value=42.0, unit="m", cumulative=False, output=True
```
florentLutz, line 87: I see what you did there 😉

src/fastoad/model_base/tests/test_flight_point.py
```diff
@@ -80,3 +80,37 @@ def test_scalarize():
...
+        "foo",
+        annotation_type=float,
+        default_value=42.0,
+        unit="slug/ft",
+        cumulative=True,
+        output=False,
+    )
+
+    assert FlightPoint.get_units()["time"] == "s"
+    assert FlightPoint.get_units()["foo"] == "slug/ft"
```
florentLutz, line 100: This doesn't seem like the recommended way of doing it now, is it worth testing ?

florentLutz (changes requested):
Thank you for the PR ! 

A few questions, remarks and changes as usual!

src/fastoad/model_base/flight_point.py
christophe-david: Here, metadata are for handling data that are not handled by the dataclass. Value and default value are handled in the "classic" Python way.

docs/documentation/mission_module/extensibility/flight_point.rst
christophe-david: Agreed. Done in 768e63f252a43106cc4e8ef0d321f3530fa44623.

src/fastoad/model_base/flight_point.py
christophe-david: Not necessary. I could have use `is_cumulative`, `is_output` and `unit` directly as fields of the metadata dict.
But the current way offers a less error-prone way to populate the metadata, and also allows having default values for theses 3 fields.

src/fastoad/model_base/flight_point.py
christophe-david: Woops.... Fixed in ea410bf211517df3bd007729274aad531c67f9f1.

src/fastoad/model_base/flight_point.py
christophe-david: Typically, it allows setting a target flight point in a segment that would be, e.g., 3000m above the start flight point.

src/fastoad/model_base/flight_point.py
christophe-david: 👍 
Fixed in a8495dfd0e9407b48464a4e65e2ec285a6e9c1bd

src/fastoad/model_base/flight_point.py
christophe-david: Actually, it is a getter :)
But if you talk about using properties, the answer is no, because here we have a class method.

src/fastoad/model_base/flight_point.py
christophe-david: Actually, we have to hack a little the dataclass mechanism for allowing the extensibility of `FlightPoint`.
Dataclasses are not designed for later addition of fields, so I had to do trial and error/debug to have it working.

Dataclasses generate their own `__init__`, and I found that a second "`dataclass` decoration" (i.e. calling `dataclass(cls)` is not enough for having an entirely new `__init__`).

I also found that after `__init__` generation, the class attributes, initially declared with `field(...)` become declared with only the default value (data in `field(...)` are transferred in `Field` instances we can access through `dataclasses.fields()`). So I had to redo a proper declaration to avoid losing all previously declared metadata during the next "`dataclass` decoration".

If you want to better understand what's going on, do not hesitate to play with the code. Removing some line and running the tests should show you the reason for the line (and if all runs smoothly, I'm interested in knowing that 😉).

src/fastoad/model_base/tests/test_flight_point.py
christophe-david: Absolutely. I had to keep `get_units()` working because it is still used at 2 locations.

src/fastoad/model_base/flight_point.py
florentLutz: I'm still not sure I understand but we can discuss that face to face, it doesn't prevent the approval of the PR

florentLutz (commented):
I still have one small comment on the way get_units is used

florentLutz (approved):
All questions answered. Thanks for the PR

PR #560, opened 2024-09-16
Code cleanup

christophe-david:
This PR aims at removing most of the warning messages triggered by FAST-OAD-core.

Any remaining warnings either come from third-party libraries, or require more attention to get rid of.


src/fastoad/_utils/arrays.py
florentLutz, line 1: I've never paid attention to that but we have one utils file per type ?

src/fastoad/_utils/pandas.py
```diff
@@ -0,0 +1,38 @@
...
+
+import pandas as pd
+from packaging.version import Version
+
+# DataFrame.applymap() is deprecated since pandas 2.1.0, in favor of DataFrame.map()
+# We get the correct one here once and for all.
+if Version(pd.__version__) >= Version("2.1"):
+    MAP_METHOD = pd.DataFrame.map
+else:
+    MAP_METHOD = pd.DataFrame.applymap
```
florentLutz, line 27: What makes you choose that way of doing things rather than setting pandas min version at 2.1 and only using applymap ?

src/fastoad/_utils/sellar/function_f.py
```diff
@@ -35,8 +35,8 @@ def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
         """Functions computation"""
 
         z2 = inputs["z"][1]
-        x1 = inputs["x"]
-        y1 = inputs["y1"]
-        y2 = inputs["y2"]
+        x1 = inputs["x"].item()
+        y1 = inputs["y1"].item()
+        y2 = inputs["y2"].item()
```
florentLutz, line 40: Two questions on that:
- Why do we need to use .items() now ?
- Shouldn't that be done in all the FAST-OAD components then ?

src/fastoad/_utils/strings.py
```diff
@@ -60,20 +59,28 @@ def get_float_list_from_string(text: str):
...
+        warnings.filterwarnings("error", category=DeprecationWarning)
 
-    try:
-        value2 = np.fromstring(text_value, dtype=float, sep=",").tolist()
-    except ValueError:
-        pass
+        try:
+            value1 = np.fromstring(text_value, dtype=float, sep=" ").tolist()
+        except (ValueError, DeprecationWarning):
+            value1 = []
```
florentLutz, line 73: That's one way of suppressing warnings 😅

src/fastoad/models/performances/mission/openmdao/tests/test_with_custom_segment.py
```diff
@@ -36,7 +36,7 @@ def cleanup():
 
 @RegisterSegment("test_segment_A")
 @dataclass
-class TestSegment(AbstractFlightSegment):
+class SegmentForTest(AbstractFlightSegment):
```
florentLutz, line 39: Why the name change ?

src/fastoad/models/performances/mission/routes.py
```diff
@@ -74,7 +75,7 @@ def cruise_distance(self):
 
     @cruise_distance.setter
     def cruise_distance(self, cruise_distance):
-        self.cruise_segment.target.ground_distance = float(cruise_distance)
+        self.cruise_segment.target.ground_distance = scalarize(cruise_distance)
```
florentLutz, line 78: Is it correct to proceed this way ? Because if for some reasons cruse_distance is an array of more than one element, you'll get an array in return while it looks like you need a float.

src/fastoad/openmdao/tests/test_utils.py
```diff
@@ -15,13 +15,18 @@
...
 from .openmdao_sellar_example.disc1 import Disc1
 from .openmdao_sellar_example.disc2 import Disc2
 from .openmdao_sellar_example.functions import FunctionF, FunctionG1, FunctionG2
 from .openmdao_sellar_example.sellar import SellarModel
 from .._utils import get_unconnected_input_names
 
+pytestmark = pytest.mark.filterwarnings(
+    "ignore:Call to deprecated function \(or staticmethod\) get_unconnected_input_names"
+)
+
```
florentLutz, line 29: That works for the whole file ?

tests/integration_tests/oad_process/data/CeRAS01_legacy_breguet_result.xml
```diff
@@ -1,6 +1,6 @@
 <!--
-  ~ This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
-  ~ Copyright (C) 2020  ONERA & ISAE-SUPAERO
+  ~ This file is part of FAST-OAD_CS25
```
florentLutz, line 2: FAST-OAD_CS25 ?

tests/integration_tests/oad_process/data/CeRAS01_legacy_breguet_result.xml
```diff
@@ -1,6 +1,6 @@
...
+  ~ Copyright (C) 2022 ONERA & ISAE-SUPAERO
```
florentLutz, line 3: 2022 ?

tests/integration_tests/oad_process/data/CeRAS01_notebooks.xml
```diff
@@ -1,5 +1,5 @@
 <!--
-  ~ This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
+  ~ This file is part of FAST-OAD_CS25
```
florentLutz, line 2: Same remark here 😅

src/fastoad/_utils/arrays.py
christophe-david: ...or per library. No golden rule here. I simply think it's clearer this way.

src/fastoad/_utils/pandas.py
christophe-david: Actually, it's "map" that has to be used from pandas 2.1.
We could do as you say, but I just wanted to keep a broad compatibility range. Some user could have a reason to stick to pandas 1.x, and want to use FAST-OAD anyway...
Of course, we won't be able to keep compatibility with old packages indefinitely, but for this one, I thought it was worth the pain.
Maybe we'll need some day to keep track of these parts of code that ensure compatibility with old versions of Python or any library, to be sure to get rid of them when the support is no longer needed.
Surely, there is somewhere a tool that does already that...

florentLutz: > Actually, it's "map" that has to be used from pandas 2.1.
Yes, sorry for the oversight
> Surely, there is somewhere a tool that does already that...
I'm sensing a sarcasm 😅

src/fastoad/_utils/sellar/function_f.py
christophe-david: Actually, I decided to do these changes in the mission module (in a pending branch), to ensure working with floats instead of 1-element arrays. The latter needed special attention before writing csv file, plus more CPU time for arithmetical operations.
The "obvious" way to have floats would be `[0]`, but numpy provides `.item()` that has some advantages, like ensuring the target is actually a 1-element array, and working alike on `np.array([0.])` and `np.array(0.)` (though the latter is not that useful in our context)

src/fastoad/_utils/strings.py
christophe-david: There is more to it. The lines above (context manager, and warning filter) are important.

src/fastoad/models/performances/mission/openmdao/tests/test_with_custom_segment.py
christophe-david: pytest considers `Test*` classes as [items to be tested](https://docs.pytest.org/en/7.1.x/getting-started.html#group-multiple-tests-in-a-class) (and probably we should have used this feature in some of our tests)

src/fastoad/models/performances/mission/routes.py
christophe-david: Well, if `cruise_distance` is an array, I want the program to crash because there is no reason for that.
Actually, the best option here would be to have `scalarize` raise an error for a multiple-element array. But I want `scalarize` to be fast, so I'm not fond of adding a new `if` in it...

src/fastoad/openmdao/tests/test_utils.py
christophe-david: yep

src/fastoad/models/performances/mission/routes.py
florentLutz: Can't the try ... except ... be used for that ?

tests/integration_tests/oad_process/data/CeRAS01_legacy_breguet_result.xml
christophe-david: Yep. To have our integration tests working with last version of FAST-OAD-CS25, I copied the xml from the CS25 repo. Strictly speaking, keeping the CS25 copyright is the right move, I think.
Anyway, given we own the 2 repos, this is of very little importance.

florentLutz (approved):

src/fastoad/models/performances/mission/routes.py
christophe-david: Actually, it's simpler than that 😅. I just have to do: 
```suggestion
        self.cruise_segment.target.ground_distance = np.asarray(cruise_distance).item()
```
 `scalarize()` is here to process the generic case where the input is "authorized" to be an n-elements array. It simply does not apply here.

src/fastoad/_utils/pandas.py
christophe-david: Not at all a sarcasm. I didn't dig the subject, but generally in software development, when a problem is generic enough, someone(s) already proposed a solution.

PR #579, opened 2024-11-25
Minor fixes in mission module

christophe-david:
This PR addresses 2 small problems:
 - lift value was generally not computed (knowing CL can be enough), though the column is always in csv files.
 - in ClimbAndCruise segment, the target of inner climb segment could be set without guarantee the climb segment was not None.

Additionally, some refactoring has been done to have CL and CD computed in only one location. For this purpose, abstract classes `AbstractLiftFromWeightSegment` and `AbstractLiftFromAoASegment` have been created, in the same purpose as `AbstractManualThrustSegment` and `AbstractRegulatedThrustSegment`.

BTW, the `load_factor` attribute has been added to `AbstractLiftFromWeightSegment`. It's an easy addition that allows to simulate segments with load factors different from 1.

### DIFF
--- a/src/fastoad/models/performances/mission/segments/registered/altitude_change.py
+++ b/src/fastoad/models/performances/mission/segments/registered/altitude_change.py
@@ -24,13 +24,16 @@ from fastoad.models.performances.mission.exceptions import FastFlightSegmentInco
 from fastoad.models.performances.mission.segments.base import (
     RegisterSegment,
 )
-from fastoad.models.performances.mission.segments.time_step_base import AbstractManualThrustSegment
+from fastoad.models.performances.mission.segments.time_step_base import (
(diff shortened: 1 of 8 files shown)


esnguyenvan (approved):
Thanks for the clean-up. Seems all good to me.

christophe-david:
> Thanks for the clean-up. Seems all good to me.

Sorry, just realized `lift_equals_weight` should have been a class attribute, not an instance one. And BTW, I consider now inheritance looks a better option to handle that. Stay tuned...

esnguyenvan (approved):
Seems good to me thanks for cleaning this up, it will solve additional issues. I would suggest to complete the documentation of the hold segment with the new load factor feature.

PR #602, opened 2025-02-12
Upgrade Github actions test to ubuntu-latest

### LINKED ISSUE #601: Migrate GitHub Actions to Ubuntu 22.04
GitHub will fully deprecate the Ubuntu 20.04 runner image by April 1, 2025. To prevent workflow failures, we need to update our GitHub Actions workflows to use ubuntu-22.04.

Required Actions:

- Identify all workflows using ubuntu-20.04.
- Update YAML files to use ubuntu-22.04 (or ubuntu-latest if preferred).
- Test workflows to ensure compatibility.

Refer to the official GitHub announcement for details: [Runner Image Deprecation Notice](https://github.com/actions/runner-images).

enricostragiotti:
## Summary of Changes
This PR initially aimed to update GitHub Action tests to use **Ubuntu latest**, but additional modifications were required to achieve this.  

The main challenge was that `mpi4py` supports multiple MPI implementations and defaults to different ones based on the operating system. Additionally, for reasons that are still unclear to me, test results varied depending on the chosen implementation. After multiple tests, the following configuration was set:  

- **Windows** → `Microsoft MPI`  
- **Linux/macOS** → `OpenMPI`  

With this setup, all tests now pass without modifying the test code. Previously, macOS tests ran without MPI, but by explicitly selecting OpenMPI, they now work correctly, so MPI tests are now enabled for macOS.  

Other modifications included:  
- Cleaning up parts of the code.  
- Adding comments in the `CalcRunner` class for better clarity.  
- Running MPI only for tests that actually use it, allowing other tests to run with `multiprocessing`.  To do that, I added a pytest marker to tag MPI-based tests and run them using `mpiexec`.  

## Related Issues

Closes #601 

## Additional Context

See here : [Original issue](https://github.com/actions/runner-images/issues/11101)


.github/workflows/tests.yml
```diff
@@ -47,16 +50,17 @@ jobs:
...
-
-        # Runners that won't send coverage report are run without coverage overhead
+      
+      ## No MPI tests
+      # Runners that won't send coverage report are run without coverage overhead
       - name: Unit tests without coverage
-        if: ${{ (runner.os != 'macOS') && ( runner.os != 'Linux' || matrix.python-version != '3.10' )}}
-        run: poetry run mpiexec -n 1 pytest src
+        if: ${{( runner.os != 'Linux' || matrix.python-version != '3.10' )}}
+        run: poetry run pytest src -m "not mpi"
```
florentLutz, line 58: How does this works ? Because I see in the pyproject.toml that you define the `mpi` marker, so is this interpreted as everything other than mpi (like a logical not) or is there also a `not mpi` marker I missed ?

.github/workflows/tests.yml
```diff
@@ -47,16 +50,17 @@ jobs:
           poetry run ruff check .           # Check coding rules
           poetry run ruff format --check .  # Check format and import sorting
         shell: bash
-
-        # Runners that won't send coverage report are run without coverage overhead
+      
+      ## No MPI tests
```
florentLutz, line 54: I suggest `Tests with MPI` and `Tests without MPI`

src/fastoad/cmd/calc_runner.py
```diff
@@ -158,6 +158,10 @@ def run_cases(
         if max_workers == -1:
             max_workers = max_proc - 1
         elif max_workers is not None:
+            if max_workers > max_proc:
+                _LOGGER.info(
+                    'Asked for "%d" workers, but only "%d" available.', max_workers, max_proc
+                )
```
florentLutz, line 164: Maybe here, that you thus set the number of workers to the max number ?

florentLutz (commented):
Thanks for the quick fix on this critical issue. One question and a few suggestion but otherwise it's all good to go

.github/workflows/tests.yml
enricostragiotti: No, by putting "not mpi" you are launching all the tests that are not marked MPI

.github/workflows/tests.yml
enricostragiotti: Yes ! My English is bad ahah

src/fastoad/cmd/calc_runner.py
enricostragiotti: Yes, this is more explicit and easier to understand for the user

enricostragiotti: I modified it into a warning in [e2a31fb](https://github.com/fast-aircraft-design/FAST-OAD/pull/602/commits/e2a31fb9f6af8c8a022ab20e6bc44aca949114e3)

PR #605, opened 2025-02-17
Set a timeout limit on GitHub Action's pipeline

### LINKED ISSUE #604: Add a maximum timeout time for tests
When dealing with #601, we noticed that some of the workflows could get stuck and run indefinitely. We should add a limit to that in order to avoid useless use of resources.

Github action permit the use of the `timeout-minutes:` keyword. We should set it to something reasonable, like 25 min. The actual runtime of test is around 15 min.

enricostragiotti:
## Summary of Changes
Added a timeout for tests and watchman tests to avoid improper usage of resources.

## Related Issues

Closes #604 

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,6 +24,7 @@ jobs:
         - { python-version: "3.9", os: "macos-latest" }
         include:
         - { python-version: "3.9", os: "macos-13" }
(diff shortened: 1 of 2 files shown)


esnguyenvan (approved):
I'm fine with this addition

PR #609, opened 2025-03-11
Ruff newrules UP RET FBT FA PTH

enricostragiotti:
## Summary of Changes  

This is the first PR introducing Ruff-based linting rules to the project to solve #610 . Currently, we rely on Codacy to enforce code quality by preventing merges when issues are detected. However, Codacy only runs checks online, requiring developers to push their code and wait for feedback before making adjustments.  

The goal of this PR is to integrate Ruff checks into our pre-commit hooks, allowing developers to check their code locally before pushing. This avoids unnecessary uploads, reduces feedback delays, and ensures that developers cannot bypass rules simply by configuring Codacy to ignore them. Instead, any ignored rule must be explicitly marked with a `noqa` directive in the code, along with a comment explaining the reason, or disabled globally in the Ruff configuration.  

### Introduced Ruff Rules  

This first batch enforces the following Ruff rules, which we agreed are important:  

- **UP (pyupgrade)**: Ensures compatibility with modern Python versions by replacing outdated syntax. [Documentation](https://beta.ruff.rs/docs/rules/#pyupgrade-up) or [orignial repo](https://github.com/asottile/pyupgrade) 
- **RET (return)**: Enforces best practices for return statements. [Documentation](https://beta.ruff.rs/docs/rules/#return-ret)  
- **FBT (flake8-boolean-trap)**: Prevents common issues with boolean function arguments. [Documentation](https://beta.ruff.rs/docs/rules/#flake8-boolean-trap-fbt)  
- **FA (flake8-annotations)**: Ensures function annotations are properly used. [Documentation](https://beta.ruff.rs/docs/rules/#flake8-annotations-fa)  
- **PTH (flake8-use-pathlib)**: Encourages the use of `pathlib` over `os.path`. [Documentation](https://beta.ruff.rs/docs/rules/#flake8-use-pathlib-pth)  

### Additional Changes  

- **Separated Ruff configuration**:  
  The Ruff configuration (`ruff.toml`) has been moved out of `pyproject.toml` to keep things clean and readable, as the configuration is expected to grow over time.  

- **Ensuring the correct Ruff version in pre-commit**:  
  The pre-commit hook now runs Ruff using Poetry to ensure the correct project-specific version is used, rather than relying on a globally installed version. This prevents potential issues, particularly in VS Code.  
  - Related issue: [VS Code Python extension bug](https://github.com/microsoft/vscode-python/issues/10165)  

- **Modification of `_is_shape_by_conn` to comply with new typing rules**:  
  The function [ _is_shape_by_conn(key, segment_class) -> bool:](https://github.com/fast-aircraft-design/FAST-OAD/blob/d04b8439644671db978f4d7c8a8ba4524286780b/src/fastoad/models/performances/mission/mission_definition/mission_builder/structure_builders.py#L228) has been modified in compliance with the new typing rules enforced by Ruff (especially problems with ClassVar and lazy type evaluation introduced by using builtin types).

## New Dependencies

updated ruff to 0.9.9

## Checklist

## Additional Context

First batch of new rules, next batch is coming once these modifications are accepted.

### DIFF
--- a/src/fastoad/models/performances/mission/mission_definition/mission_builder/structure_builders.py
+++ b/src/fastoad/models/performances/mission/mission_definition/mission_builder/structure_builders.py
@@ -20,9 +20,8 @@ be transformed into a Python implementation.
 
 from abc import ABC, abstractmethod
 from copy import deepcopy
-from dataclasses import InitVar, dataclass, field
+from dataclasses import InitVar, dataclass, field, fields
 from itertools import chain
-from typing import ClassVar, get_args, get_origin, get_type_hints
 
 import numpy as np
 
@@ -230,44 +229,14 @@ class AbstractStructureBuilder(ABC):
         Here variables that are expected to be arrays or lists in the provided segment class are
         attributed the "shape_by_conn=True" property.
         Check if a field in `segment_class` is a list or NumPy array.
-        Works with `from __future__ import annotations` and properly handles the ClassVars
-        defined in `segment_class`.
         """
-
+        segment_fields = [fld for fld in fields(segment_class) if fld.name == key]
         try:
-            # Resolve type annotations to handle cases where they are stored as strings due to
-            # `from __future__ import annotations`. This ensures we work with real types.
-            # TODO once https://peps.python.org/pep-0649/ is implemented, get_type_hints should
-            # not be necessary anymore.
-            type_hints = get_type_hints(segment_class)
-
-            if key not in type_hints:
-                return False  # If the field does not exist, return False immediately.
-
-            field_type = type_hints[key]
-
-            # Handle ClassVar, which wraps types but is not itself a valid type for `issubclass()`.
-            origin = get_origin(field_type)
-            if origin is ClassVar:
-                field_type = get_args(field_type)[0]  # Extract the real type inside ClassVar.
-                origin = get_origin(field_type)  # Re-evaluate origin in case it's a generic type.
-
-            # If the field is explicitly declared as a `list` (e.g., `list[int]`), return True.
-            if origin is not None and origin is list:
-                return True
-
-            # Check if the field is a subclass of `list` or `np.ndarray`.
-            # This handles cases where the type is directly `list` or `np.ndarray` without generics.
-            if isinstance(field_type, type):
-                try:
-                    return issubclass(field_type, (list, np.ndarray))
-                except TypeError:
-                    pass  # Some types cannot be used in `issubclass()`, so we ignore errors.
-
-            return False  # If no conditions matched, return False.
-
-        except (TypeError, NameError):
-            return False  # Catch errors from missing imports or invalid type hints.
+            return len(segment_fields) == 1 and issubclass(
+                segment_fields[0].type, (list, np.ndarray)
+            )
+        except TypeError:
+            return len(segment_fields) == 1 and isinstance(segment_fields[0], (list, np.ndarray))
 
     def _process_polar(self, structure):
         """


src/fastoad/models/performances/mission/segments/registered/takeoff/__init__.py
```diff
@@ -16,9 +16,8 @@
...
 #  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-
-# flake8: noqa
-
 # With these imports, importing only the current package ensures to have all
 # these segments available when interpreting a mission input file
 from . import end_of_takeoff, rotation, takeoff
+
+__all__ = ["end_of_takeoff", "rotation", "takeoff"]
```
esnguyenvan, line 23: Not sure I understand the need for this declaration. It seems to me a sort of duplication of line 21. Can you point me toward an explanation?

src/fastoad/models/performances/mission/mission_definition/mission_builder/structure_builders.py
```diff
@@ -227,11 +228,40 @@ def _parse_inputs(
...
     def _is_shape_by_conn(key, segment_class) -> bool:
         """
-        Here variables that are expected to be arrays or lists in the provided segment class are
-        attributed the "shape_by_conn=True" property.
+        Check if a field in `segment_class` is a list or NumPy array.
+        Works with `from __future__ import annotations` and handles ClassVar properly.
         """
-        segment_fields = [fld for fld in fields(segment_class) if fld.name == key]
-        return len(segment_fields) == 1 and issubclass(segment_fields[0].type, (list, np.ndarray))
+        # Retrieve type annotations using get_type_hints().
```
esnguyenvan, line 234: It seems to me that the function does more than it is expected to. Wouldn't it be possible to deactivate the typing rule locally? Especially since the exact type is not necessary here, we just want to know if the field is a single element or not.

esnguyenvan (changes requested):
Hey, thanks for this important work. Overall I agree with the changes, it makes the code nice and consistent. I have a few questions and remarks as well.

esnguyenvan:
In addition to the previous comments, the changes brought by this PR will make the following integration tests of CS25 break:

 - FAILED tests/integration_tests/oad_process/test_oad_process.py::test_api_eval_breguet - AttributeError: 'bool' object has no attribute 'lower'
 - FAILED tests/integration_tests/oad_process/test_oad_process.py::test_api_eval_mission - AttributeError: 'bool' object has no attribute 'lower'
 - FAILED tests/integration_tests/oad_process/test_oad_process.py::test_api_optim - AttributeError: 'bool' object has no attribute 'lower'

src/fastoad/cmd/calc_runner.py
```diff
@@ -11,6 +12,7 @@
 #  GNU General Public License for more details.
 #  You should have received a copy of the GNU General Public License
 #  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from __future__ import annotations
```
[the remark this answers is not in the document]
enricostragiotti: https://docs.astral.sh/ruff/rules/future-required-type-annotation/

from __future__ import annotations is needed because since 3.10 you can  convert Optional[str] to str | None. But if we want compatibility with Python 3.9 you need to add this line

src/fastoad/models/performances/mission/segments/registered/takeoff/__init__.py
enricostragiotti: Explication here 
https://docs.astral.sh/ruff/rules/unused-import/
the idea is that just importing creates unused imports. putting them to all declares that they should be imported in a wildcard import and gives them a "reason" to stay in the init.

More info here
https://peps.python.org/pep-0008/#public-and-internal-interfaces
https://stackoverflow.com/questions/44834/what-does-all-mean-in-python

src/fastoad/io/configuration/configuration.py
```diff
@@ -320,7 +323,7 @@ def _make_options_local(
         self,
         structure: dict,
         new_root_path: Path,
-        local_path: Path = Path("."),
+        local_path: Path = Path("."),  # noqa: PTH201
```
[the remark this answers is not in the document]
enricostragiotti: my bad, I forgot that :)

src/fastoad/models/performances/mission/mission_definition/mission_builder/structure_builders.py
enricostragiotti: The previous implementation of `_is_shape_by_conn()` directly accessed `segment_fields[0].type` and used `issubclass()`, which caused issues when `from __future__ import annotations` was used. This made annotations lazy (evaluated only at runtime), storing them as strings instead of actual types, requiring `get_type_hints()` to properly resolve them. Additionally, the function did not handle `ClassVar`, which cannot be used in `issubclass()` and would raise a `TypeError`. The updated implementation first retrieves type hints using `get_type_hints()`, ensuring compatibility with future annotations behavior. It then unwraps `ClassVar` to extract the actual type before performing type checks. Using `typing.get_origin()`, it also correctly handles generics like `list[int]`. Finally, it includes exception handling to avoid `TypeError` in cases where `issubclass()` is not applicable. These changes make the function more robust, ensuring it works with both explicit and future-proofed type annotations.

enricostragiotti: I know that is not very easy to read, but it is to comply with recent modification. do you think that I should add more comments in the code?

src/fastoad/cmd/calc_runner.py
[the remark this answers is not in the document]
esnguyenvan: OK thanks. I understand that the import may be removed when dropping 3.9 then?

src/fastoad/models/performances/mission/segments/registered/takeoff/__init__.py
esnguyenvan: OK thanks for the explanations

src/fastoad/models/performances/mission/mission_definition/mission_builder/structure_builders.py
esnguyenvan: I understand the function should be more robust but I fear it'll be quite a challenge to maintain (and I understood that it will be changed in the future).

The complexity deployed seems quite high for the final objective (check if the field is a single or not). The question is, do we need such complexity here?

If yes, then yes indeed I need more comment.

If not, maybe it's possible to deactivate the ruff rule for now and activate it in the future when 3.9 will be dropped and the code will be fully compatible with PEP 604

enricostragiotti:
thanks @esnguyenvan for the remarks ! I completely reworked the logic of the _is_shape_by_conn method, just by handling in a try except the exception that we had because of the use of from __future__ import annotations and the lazy type evaluation.
I also checked that now all the CS25 tests pass. Could you tell me when you do it if the RTA tests pass as well? Thank you very much !

esnguyenvan (approved):
Thanks for the rework, it's quite nice! It's all good for me here. It's all good with the RTA repo

src/fastoad/models/performances/mission/mission_definition/mission_builder/structure_builders.py
enricostragiotti: reworked and simplified in https://github.com/fast-aircraft-design/FAST-OAD/pull/609/commits/9d4b868df5b1f512dfd0e6ee8131214a06c45fc6

enricostragiotti:
Perfect! Let's wait for @ScottDelbecq review then! thanks Éric!

PR #616, opened 2025-03-19
Fix inconsistent behaviour of the get_val method when not providing units of measure

### LINKED ISSUE #606: Inconsistent Return Type in Variable.get_val(new_units=None)
**Describe the bug**
The method Variable.get_val(new_units=None) returns a list, while Variable.get_val(new_units="UnitOfMeasure") returns a scalarized value. This inconsistency leads to unexpected behavior:

- If the variable has no units, get_val(new_units=None) returns a list with one element, even if it's a scalar.
- If the variable has units, get_val(new_units="UnitOfMeasure") returns a scalarized value, even if the original value is a vector.

**To Reproduce**
val1 = variable.get_val(new_units=None)  # Returns a list
val2 = variable.get_val(new_units="m")  # Returns a scalarized value

**Expected behavior**
- Consistency in return types.
- If the variable is a scalar, get_val() should return a scalar, regardless of the units.
- If the variable is a vector, get_val() should return a vector, regardless of the units.

enricostragiotti:
## Summary of Changes

Fixed the consistency of the `get_val` method. This PR introduces a small breaking change: previously, if the FAST-OAD output was a 1D iterable (i.e., a scalar), `get_val` would return an iterable, while requesting the same value with unit conversion would return a scalar. This inconsistency has been fixed, so it is no longer necessary to do `prob.get_val("var_name")[0]` to extract the scalar value. If the exact raw value from the FAST-OAD output is needed, it can still be accessed using `prob["var_name"]`.

## Related Issues

closes #606 
## Checklist

- [x] Does this PR introduce a breaking change?

### DIFF
--- a/src/fastoad/openmdao/tests/test_variables.py
+++ b/src/fastoad/openmdao/tests/test_variables.py
@@ -14,7 +14,6 @@ Module for testing VariableList.py
 #  You should have received a copy of the GNU General Public License
 #  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from collections.abc import Iterable
 from pathlib import Path
 from typing import List
 
@@ -850,13 +849,14 @@ def test_get_val():
     vars["qux"] = {"value": np.array([1.0, 2.0]), "units": "m"}
     vars["quux"] = {"value": [[1.0, 2.0], [2.0, 3.0]], "units": "m"}
     data = vars["bar"].get_val()
-    assert not isinstance(data, Iterable) and not isinstance(data, np.ndarray)
+    assert not isinstance(data, list) and not isinstance(data, np.ndarray)
     assert_allclose(data, 1, rtol=1e-3, atol=1e-5)
-    data = vars["baq"].get_val()
-    assert not isinstance(data, Iterable) and not isinstance(data, np.ndarray)
     units = "km"
     data = vars["bar"].get_val(new_units=units)
     assert_allclose(data, 1e-3, rtol=1e-3, atol=1e-5)
+    data = vars["baq"].get_val()
+    assert not isinstance(data, list) and not isinstance(data, np.ndarray)
+    assert_allclose(data, 1, rtol=1e-3, atol=1e-5)
     with pytest.raises(TypeError):
         data = vars["foo"].get_val(new_units=units)
     data = vars["baz"].get_val(new_units=units)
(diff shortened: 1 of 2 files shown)


enricostragiotti:
@esnguyenvan Thanks for the tips! i will rework the doc! Concerning the modification to do in CS25, I will draft a pr and we'll use it as soon as we release a new version of fastoad core with this specific PR. Is it ok for you?

src/fastoad/openmdao/tests/test_variables.py
```diff
@@ -842,13 +843,17 @@ def test_get_variables_from_problem_sellar_with_promotion_and_connect():
...
     vars = VariableList()
     vars["bar"] = {"value": 1.0, "units": "m"}
+    vars["baq"] = {"value": np.array([1.0])}
     vars["foo"] = {"value": 1.0, "units": "m**2"}
     vars["baz"] = {"value": [1.0, 2.0], "units": "m"}
     vars["bat"] = {"value": (1.0, 2.0), "units": "m"}
     vars["qux"] = {"value": np.array([1.0, 2.0]), "units": "m"}
     vars["quux"] = {"value": [[1.0, 2.0], [2.0, 3.0]], "units": "m"}
     data = vars["bar"].get_val()
+    assert not isinstance(data, Iterable) and not isinstance(data, np.ndarray)
```
florentLutz, line 853: An ndarray is already an iterable no ? So the first clause is redundant no ? And just a question, what would happen with a one dimensionnal tuple ?

florentLutz (changes requested):
Redundant clause in the tests

src/fastoad/openmdao/tests/test_variables.py
florentLutz: Thanks for the change, but now that I think about it does it really make sense to test for what type it is not ? Rather couldn't we just test that vars["bar"].get_val() returns 1.0 (as is done) and the same for baq ?

It's an open question. If you think it makes more sense to leave it as is I'm happy with how it is done

enricostragiotti: i completely understand your point here, it was only an additional check. this test is to be sure that we will never have a type error when using get_val and give the output so a function that works only with scalars (it is not the case for well coded function like assert_allclose for example, where scalarize is called internally).

As the return type is an important part of the method, directly testing the type makes sense in my head. It could "protect" us by type errors if someone decided to change the scalarize function or the get_val method.

florentLutz (approved):

PR #617, opened 2025-03-20
Upgrade to openmdao 3.38

relf:
## Summary of Changes

Openmdao 3.38 introduces a breaking changes in `problem.setup()`: see https://github.com/OpenMDAO/OpenMDAO/pull/3458. 
The changes enables initialization of dynamically-shaped variables but part of the setup is now reported in the `problem.final_setup()` method which means that now some initialization occurs later. From the end-user point of view it boils down to calling `pb.final_setup()` but inner FAST-OAD variables handling needs finer control specially by using the `set_setup_status()` which allows to handle intermediate setup states.

Note: this PR is not compatible with openmdao < 3.38 hence the `^3.38` requirement

## Related Issues

#612 

## Checklist

- [x] Have you followed the guidelines in our [Contributing](https://github.com/fast-aircraft-design/FAST-OAD/wiki/Development-environment) wiki?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
- [ ] Have you added new tests to cover your changes?
- [ ] Have you made corresponding changes to the documentation? If yes, please consider providing a link to the updated documentation.
- [x] Do the existing tests pass with your changes?
- [ ] Have you commented on hard-to-understand areas of the code?
- [ ] Does this PR introduce a breaking change?

PR #618, opened 2025-03-25
Watchman tests now ignore the pyproject toml requirements

enricostragiotti:
## Summary of Changes

This small modification of the Watchman test allows the tests to install the latest version of OpenMDAO without considering the possible requirements specified in the `pyproject.toml` file. As expected, the test now fails because the latest version of OpenMDAO is still incompatible with FAST at the time of writing this PR.


esnguyenvan:
Hey!
OK it works as expected. But now that I think about it, shouldn't we add this to https://github.com/fast-aircraft-design/FAST-OAD/pull/617 ?
It can stay as a separate PR but there's no need for this change until we merge https://github.com/fast-aircraft-design/FAST-OAD/pull/617.

PR #622, opened 2025-04-03
WIP

enricostragiotti:
## Checklist

- [ ] Have you followed the guidelines in our [Contributing](https://github.com/fast-aircraft-design/FAST-OAD/wiki/Development-environment) wiki?
- [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
- [ ] Do the existing tests pass with your changes?


enricostragiotti:
I'll do a little ping here @esnguyenvan :) That will permit to go on during summer with the DOE functionality :)

src/fastoad/openmdao/variables/variable_list.py
```diff
@@ -121,16 +123,27 @@ def update(self, other_var_list: list, add_variables: bool = True):
             - if not, Variable instance from other_var_list will be added only if
               add_variables==True
 
+        The merge_metadata parameter controls how metadata is handled when replacing variables:
+            - if merge_metadata=False (default): The variable is completely overwritten with the
+              new one, and all previous metadata is lost (except for descriptions as noted above)
+            - if merge_metadata=True: Only the metadata explicitly provided in the new variable
```
esnguyenvan, line 129: From how I understand the function `update_missing_metadata`, this should be changed to:
```suggestion
            - if merge_metadata=True: Only the metadata explicitly provided in the new variable and not present in the original variable
```
Although I don't know if that was the intended behaviour.

src/fastoad/openmdao/variables/variable.py
```diff
@@ -101,21 +103,23 @@ def __init__(self, name, **kwargs):
...
-            comp = om.ExplicitComponent()
-            # get attributes
-            metadata = comp.add_output(name="a")
-
-            self.__class__._base_metadata = metadata
-            self.__class__._base_metadata["val"] = 1.0
-            self.__class__._base_metadata["tags"] = set()
-            self.__class__._base_metadata["shape"] = None
-        # Done with class attributes ------------------------------------------
+        init_metadata = kwargs.pop("init_metadata", True)
```
esnguyenvan, line 106: Why not declaring the 'init_metadata' as an optional argument of the `__init__()` ? This way we could make the difference more clearly between this parameter and those related to variable metadata. ( Plus this keyword argument would be fed to `self.metadata.update()` method )

src/fastoad/openmdao/variables/variable_list.py
esnguyenvan: Well actually I think the description is misleading. I understood it the other-way-around.

Should it be:
`if merge_metadata = True : The metadata present in the original variable and absent from the new variable will be conserved. All other metadata will be overwritten by the ones in new variable.`

src/fastoad/openmdao/tests/test_variables.py
```diff
@@ -127,6 +128,176 @@ def test_variables(with_dummy_plugin_2):
...
+    no_meta_var2 = Variable("var2", val=25.0, desc="New description", init_metadata=False)
+    no_meta_var_list = VariableList([no_meta_var1, no_meta_var2])
+
+    # Update with merge_metadata=True
+    var_list.update(no_meta_var_list, merge_metadata=True)
+
+    # Check results
+    assert var_list["var1"].value == 15.0
+    assert var_list["var1"].units == "ft"
+    assert var_list["var1"].description == "Description for var1"  # Preserved
```
esnguyenvan, line 294: Shouldn't you test on another metadata as the standard behaviour is to preserve the description, unless merge_metadata is called?

esnguyenvan (changes requested):
Hey, I mainly have questions about the descriptions that I found misleading

src/fastoad/openmdao/variables/variable_list.py
enricostragiotti: Yes is much clearer like this!

src/fastoad/openmdao/variables/variable.py
enricostragiotti: yes I agree, it would be much cleaner! Thanks! (I haven't understood what you've written inside the parenthesis)

enricostragiotti: fixed in https://github.com/fast-aircraft-design/FAST-OAD/commit/0849527225f94448aabea2e2280a0c9178b3ab81

PR #627, opened 2025-04-15
Deactivate OpenMDAO automatic reporting

### LINKED ISSUE #626: Disable OpenMDAO reporting
Everytime an OpenMDAO problem and by extension a FAST-OAD problem is created and ran, OpenMDAO creates a report which writes files. Given the hundreds of unit tests ran each time, this can quickly become annoying?

Disabling OpenMDAO reporting could be a solution since it is never used (it is rather recommend to use the output file in the .XML format).

Deleting manually the files 😅

florentLutz:
## Summary of Changes

Reinstated OpenMDAO's reporting deactivation when using FASTOADProblem (it used to work with `None `but now it needs to be set to `False`). Also disabled reporting in tests and switch OpenMDAO problem for FAST-OAD problem in the test functions when possible.

## Related Issues

This PR closes #626 

## Checklist

### DIFF
--- a/src/fastoad/_utils/sellar/tests/test_sellar.py
+++ b/src/fastoad/_utils/sellar/tests/test_sellar.py
@@ -17,7 +17,7 @@ from fastoad._utils.sellar.sellar_base import BasicSellarModel, BasicSellarProbl
 
 
 def test_sellar():
-    problem = BasicSellarProblem(BasicSellarModel(), reports=False)
+    problem = BasicSellarProblem(BasicSellarModel())
 
     problem.setup()
     problem.run_driver()
--- a/src/fastoad/_utils/sellar/tests/test_sellar.py
+++ b/src/fastoad/_utils/sellar/tests/test_sellar.py
@@ -17,7 +17,7 @@ from fastoad._utils.sellar.sellar_base import BasicSellarModel, BasicSellarProbl
 
 
 def test_sellar():
-    problem = BasicSellarProblem(BasicSellarModel())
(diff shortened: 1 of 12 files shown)


src/fastoad/_utils/sellar/tests/test_sellar.py
```diff
@@ -17,7 +17,7 @@
 
 
 def test_sellar():
-    problem = BasicSellarProblem(BasicSellarModel())
+    problem = BasicSellarProblem(BasicSellarModel(), reports=False)
```
enricostragiotti, line 20: Would it be possible to modify directly the BasicSellarProblem here instead? in a similar fashion of how we do for a FASTOADProblem

enricostragiotti (changes requested):
Thanks for that PR! The tests pass and no report is generated ad all. Finally we can get rid of all those reports! Just a minor change and we are all ok! 

PS would it be possible to rename the PR "decative reporting" into something more explicit, like "Decativate OpenMDAO automatic reporting in tests"? It will ease the writing of the release report :)

src/fastoad/_utils/sellar/tests/test_sellar.py
florentLutz: It is 😄 I actually thought this was something straight from OpenMDAO hence why I deactivated it at instantiation but now it is fixed in e2261039

enricostragiotti:
All good for me! Thanks for this very needed contribution !

Issue #597, opened 2025-02-05
Objective function not always normalised when determining the final flightpoint of a segment


esnguyenvan:
**Describe the bug**
In the mission, there is an simple optimisation algorithm to find the time at which the target is reached. This algorithm is called within a while loop with tolerance reducing at each loop. However, the quantity of interest is not normalised before calling the optimisation algorithm. In some cases, if the quantity of interest is very large (lift for large aircraft), the while loop never exits.

**To Reproduce**
The bug is very case dependent and may be difficult to reproduce.

**Expected behavior**
The while loop may be a problem as the mission can remain stuck without error message.
Normalising the quantity of interest before entering the optimisation algorithm should be sufficient to ensure robustness of the code.

PR #628, opened 2025-04-15
Fix mission target calculation

esnguyenvan:
## Summary of Changes

This PR addresses the problem described by #597.

The base time-step algorithm finds the exact time at which the target is reached with a tolerance of 1e-5 (absolute tolerance). It does so using a Newton method that iterates over the time step until the distance to target reaches 0. The Newton method is however not precise enough to reach the tolerance of 1e-5. So it is located within a while loop that gradually decreases the tolerance of the Newton solver. This is still not general enough and can lead to mission being stuck in the while loop.

The changes consist of:

- Modifying an existing test to enter the never ending while loop. The test fails if timed-out.
- Normalising the distance to target with the last_point_to_target
- Getting rid of the while loop

## Related Issues
Closes issue #597.

## New Dependencies
No new dependencies.

## Checklist

### DIFF
--- a/src/fastoad/models/performances/mission/segments/registered/tests/test_takeoff_segments.py
+++ b/src/fastoad/models/performances/mission/segments/registered/tests/test_takeoff_segments.py
@@ -11,6 +11,8 @@
 #  You should have received a copy of the GNU General Public License
 #  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import multiprocessing.pool
+
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose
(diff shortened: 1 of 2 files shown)


dimitris-glenis:
hey thanks for the nice work! I have some questions/remarks:

- Maybe it would be nice to add a threshold of iterations in your while loop in case it doesn't reach the desired tolerance.
- Have a flag on your root_scalar function to check if converged
- Why did you set the xtol on the root_scalar to 1e-6?

esnguyenvan:
Hey, thanks for the review.

About your questions:

1. Difficult to implement because the while loop applies to the segment. And each segment has a different length that can hardly be determined in advance and different time step that can be set by the user. 
2. Agreed and implemented as a warning. But not sure how to use this information at the moment.
3. The tolerance on top (for the while) is 10-5 so I set a tolerance slightly smaller to be sure that the while loop can be exited.

Hope that answers your questions.

The watchman Tests are not passing is expected because of changes in Openmdao

dimitris-glenis (approved):
Looks good to me

PR #639, opened 2025-07-09
Added more verbose logging for failed module import

### LINKED ISSUE #635: Improve error messages when registering new modules with `RegisterOpenMDAOSystem` or `RegisterSubmodel`
When adding a new submodel module and registering it using `RegisterOpenMDAOSystem` or `RegisterSubmodel`, it is currently very hard to identify what causes a failure. If there's a simple mistake (e.g. a typo in an import), the resulting error during registration provides no useful context: no filename, no line number, no traceback.

As a developer, you're left with only a generic `FastNoSubmodelFoundError`, which makes it appear like no submodel was found — when in reality, the module failed to import silently.

It would be very helpful to have a **more verbose error output** during the registration process. Ideally, this would include:

- The name/path of the module that failed to register  
- The specific error traceback or at least the exception type and message  
- Context indicating whether the failure occurred during import or registration

This would significantly improve the developer experience and make it easier to integrate and debug new components.

enricostragiotti:
## Summary of Changes

This PR improves the logging around the dynamic installation of Python packages and bundles in the FAST-OAD framework, specifically in the `_install_python_package` method run to discover the packages that can be registered by FAST-OAD.

## Related Issues

closes #635 

## Additional Context

Here is the output we obtain for the example shown in #635:

```bash
WARNING  root:_bundle_loader.py:386 ================================================================================
WARNING  root:_bundle_loader.py:387     fastoad_cs25.models.geometry.geom_components.wing.components
WARNING  root:_bundle_loader.py:388 --------------------------------------------------------------------------------
WARNING  root:_bundle_loader.py:390 
!! Exception: Error installing bundle fastoad_cs25.models.geometry.geom_components.wing.components.compute_b_50: No module named 'openmdao.ap'

  >> Full error traceback:
Traceback (most recent call last):
...

WARNING  root:_bundle_loader.py:391 --------------------------------------------------------------------------------

WARNING  root:_bundle_loader.py:390 
!! Exception: Error installing bundle fastoad_cs25.models.geometry.geom_components.wing.components.compute_planform: No module named 'openmdao.ap'

  >> Full error traceback:
Traceback (most recent call last):
...
```

and

```bash
WARNING  root:_bundle_loader.py:106 ================================================================================
WARNING  root:_bundle_loader.py:107 !!  FAST-OAD MODULE IMPORT FAILURE RECAP
WARNING  root:_bundle_loader.py:108 --------------------------------------------------------------------------------
WARNING  root:_bundle_loader.py:110   - fastoad_cs25.models.geometry.geom_components.wing.components.compute_b_50
WARNING  root:_bundle_loader.py:110   - fastoad_cs25.models.geometry.geom_components.wing.components.compute_planform
WARNING  root:_bundle_loader.py:110   - fastoad_cs25.models.geometry.geom_components.wing.components.tests.test_wing_geom_comps
WARNING  root:_bundle_loader.py:111 ================================================================================
```

Before this PR, we only get

```bash
WARNING  fastoad.module_management._bundle_loader:_bundle_loader.py:106 Failed to import module fastoad_cs25.models.geometry.geom_components.wing.components
```

with no additional info on the real root cause of the problem, forcing the dev to spend time to setup a script launching the package, or debugging.

**IMPORTANT**
In this PR the tracebacks can be quite verbose. This is especially true as you would get a traceback not only for the failing file, but also for all the files that import it (e.g. om.group). trying to overcome that would have been a little overkill for me, and I'm not bothered by the quantity of information if it is useful.


src/fastoad/module_management/_bundle_loader.py
```diff
@@ -34,7 +35,9 @@
 from .._utils.files import as_path
 from .._utils.resource_management.contents import PackageReader
 
-_LOGGER = logging.getLogger(__name__)
+_LOGGER = logging.getLogger(
+    ""
+)  # Using the root logger here just because we want less verbose outputs
```
florentLutz, line 40: I'd opt against using the root logger here for EVERY logged message of that module. the reason being for instance that I, as a user, tend to be tired of hundreds of message saying such or such module has been installed correctly so I tend to ignore those logged messages. But I'd still want warning if one was not installed correctly. 

All that to say, would it be possible to have two different loggers here :
- One for message that says everything is going well which any user could deactivate
- One for errors

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -102,8 +105,13 @@ def explore_folder(
                 "Installed bundle %s (ID %s )", bundle.get_symbolic_name(), bundle.get_bundle_id()
             )
             bundle.start()
-        for _f in failed:
-            _LOGGER.warning("Failed to import module %s", _f)
+        if failed:
+            _LOGGER.warning("=" * 80)
+            _LOGGER.warning("!!  FAST-OAD MODULE IMPORT FAILURE RECAP")
+            _LOGGER.warning("-" * 80)
```
florentLutz, line 111: So this is the recap, yes ? 

Wouldn't there be a way to make it a bit nicer ? Maybe a table with faulty file and said error ? I imagined somthing like the list_modules command.
Also by running the thing suggested in #635 I notice that as you said, the same error is actually reported twice. Once in the faulty module and once in a module which imports the faulty module. Would there be a way to show the error only once ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -361,7 +369,11 @@ def _install_python_package(self, package_name: str) -> Tuple[Set[Bundle], Set[s
         package = PackageReader(package_name)
         if package.has_error or not package.exists:
             failed.add(package_name)
+            _LOGGER.warning("=" * 80)
+            _LOGGER.warning("    %s", package_name)
+            _LOGGER.warning("-" * 80)
```
florentLutz, line 374: And this is for individul failure ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -371,8 +383,19 @@ def _install_python_package(self, package_name: str) -> Tuple[Set[Bundle], Set[s
                         try:
                             bundle = self.context.install_bundle(item_package[:-3])
                             bundles.add(bundle)
-                        except BundleException:
-                            failed.add(package_name)
+                        except BundleException as e:
```
florentLutz, line 386: I'm not sure I understand why  two separate loggings are required line 372->374 and line 386->398. Out of curiosity, could you explain it to me ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -371,8 +383,19 @@ def _install_python_package(self, package_name: str) -> Tuple[Set[Bundle], Set[s
...
+                            failed.add(item_package[:-3])
```
florentLutz, line 387: What does the -3 correspond to ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -371,8 +383,19 @@ def _install_python_package(self, package_name: str) -> Tuple[Set[Bundle], Set[s
...
+                            if not header_printed:
+                                _LOGGER.warning("=" * 80)
+                                _LOGGER.warning("    %s", package_name)
+                                _LOGGER.warning("-" * 80)
+                                header_printed = True
+                            _LOGGER.warning(
+                                "\n!! Exception: %s\n\n  >> Full error traceback:\n%s",
+                                e,
+                                traceback.format_exc(),
```
florentLutz, line 396: Is there more than one format ? If yes, have you looked into one that would give a "cleaner" look ?

florentLutz (changes requested):
A few recommandation based on user feedback, some questions to make things cleaner but mainly, I was wondering if a test shouldn't be added:
a) To have an easier way to access and tune the test
b) and to check that it captures what we want to capture
Other than that, thanks a lot for that PR, I don't know if its coicidental but a lot of PhD students have requested that feature at ISAE ^^'

src/fastoad/module_management/_bundle_loader.py
enricostragiotti: I thought a bit about that, and no, this wouldn't be easy to do. I spoke about that with @relf and @esnguyenvan and they basically said to me that this wasn't really needed and that it was like using a cannon to kill a mosquito :)

src/fastoad/module_management/_bundle_loader.py
enricostragiotti: The first if and the corresponding print is if the whole package_name is invalid (e.g. not found, not accessible, or not a Python module). The second, instead, is used when the package is valid and exists. You are then trying to recursively load its modules, but one or more specific .py files (bundles or modules) fail during installation.

src/fastoad/module_management/_bundle_loader.py
enricostragiotti: Thanks for the feedback, but I think there might be a bit of a misunderstanding.

I used the root logger mainly to avoid having overly verbose prefixes like fastoad.bundle.loader._internal.bundle_loader in every message — not to force everything into a single logger that users can turn on or off.

I’m not in favor of distinguishing between loggers for "optional" info and "critical" errors in this context. In my view, an error is an error — if a module fails to install and throws an exception, it must be logged, with the full traceback. That’s not something the user should choose to ignore.

If someone wants to avoid seeing errors for a certain folder or package, they can choose not to scan it. But once it's scanned, any failure should be reported — no silencing that.

That said, I'm open to improving how logs are structured or filtered to make them clearer without losing important information.

florentLutz: That being said, I know why it bothers me but not really other users from FAST-OAD. In FAST-OAD_CS23 we tend to log at the bottom most level so we get every message hence why we feel the need to disable them.

If you feel like my explanations aren't clear we can discuss that in today's developers meeting

pyproject.toml
```diff
@@ -92,6 +92,7 @@ importlib-metadata = { version = "^4.2", python = "<3.10" }
 mpi4py = [
     { version = "^4.0.2", optional = true },
 ]
+rich = "^14.0.0"
```
florentLutz, line 95: Couldn't we do it with the existing package (like the one used for `list_modules`) ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -35,6 +39,25 @@
 from .._utils.resource_management.contents import PackageReader
 
 _LOGGER = logging.getLogger(__name__)
+_LOGGER.propagate = False
```
florentLutz, line 42: What does that do  ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -35,6 +39,25 @@
...
+_LOGGER.setLevel(logging.WARNING)
+
+# Create RichHandler with custom formatter (only message)
+rich_handler = RichHandler(
```
florentLutz, line 46: Isn't there some sort of convention that every variable define outside of classes or function should be upper case ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -353,33 +376,71 @@ def _install_python_package(self, package_name: str) -> Tuple[Set[Bundle], Set[s
...
         elif package.is_package:
-            # It is a package, let's explore it.
+            header_printed = False  # to print error header once per package
+
             for item in package.contents:
                 item_package = ".".join([package_name, item])
+
                 if "." in item:
-                    # A file. Considered only if it is a Python file. Ignored otherwise.
+                    # It's a file
```
florentLutz, line 396: I liked the original comments though I missed them on my first review round ^^' Why delete them ?

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -353,33 +376,71 @@ def _install_python_package(self, package_name: str) -> Tuple[Set[Bundle], Set[s
...
+                    # It's a subpackage
                     sub_bundles, sub_failed = self._install_python_package(item_package)
                     bundles.update(sub_bundles)
                     failed.update(sub_failed)
 
         return bundles, failed
 
+    @staticmethod
+    def _log_failed_modules(failed_modules: list[str]):
+        table = Table(
```
florentLutz, line 420: Isn't there a way to print the full path to the failing module ? Again, like what is done in the `list_module `method ?

src/fastoad/notebooks/02_From_Python_to_FAST-OAD/05_FAST-OAD.ipynb
florentLutz, line 1: This is not really a change, it's just because another env was used to run the notebook. IMO you should discard it ^^

src/fastoad/module_management/_bundle_loader.py
enricostragiotti: I now understand your concern! In working on this PR, I’ve learned quite a bit about Python loggers, I’m definitely not an expert, but I think I get the basics now. If I understand correctly, there are multiple logging levels, and by default not all of them are visible at runtime.

Typically, only messages at the WARNING level and above are shown (i.e. WARNING, ERROR, and CRITICAL), while lower levels like INFO and DEBUG are hidden unless explicitly enabled.

So yes, we don’t need multiple loggers to achieve that behavior: it’s already built into the logging system and used by FAST.

src/fastoad/module_management/_bundle_loader.py
enricostragiotti: For me this is only if they are constants, no? here i just need to use it on this specific module

src/fastoad/module_management/_bundle_loader.py
enricostragiotti: The full **absolute** path? do you think that would be useful and not very verbose? You would want maybe a table like:

Name of the module | absolute path

?

src/fastoad/notebooks/02_From_Python_to_FAST-OAD/05_FAST-OAD.ipynb
enricostragiotti: yes definetly, my bad :)

pyproject.toml
enricostragiotti: Rich is what enables traceback to be easy to understand. I think it is very important to have it! Have a look here:
https://www.willmcgugan.com/blog/tech/post/better-python-tracebacks-with-rich/
or
https://mathspp.com/blog/til/043

src/fastoad/module_management/_bundle_loader.py
florentLutz: True but then, why are you considering that ther `rich_handler `is not a constant while the `_LOGGER `is ? ^^

pyproject.toml
florentLutz: I'm convinced. 😄

src/fastoad/module_management/_bundle_loader.py
florentLutz: That they don't affect you mean ?

enricostragiotti: yes sorry 😮

src/fastoad/module_management/_bundle_loader.py
enricostragiotti: Done in https://github.com/fast-aircraft-design/FAST-OAD/pull/639/commits/bdfcd4fc0038d878949add489a9f1d0622628211

src/fastoad/module_management/_bundle_loader.py
```diff
@@ -347,39 +372,88 @@ def _get_service_references(
...
         bundles = set()
-        failed = set()
+        failed = {}  # dict {module_name: full_path}
 
         package = PackageReader(package_name)
+
+        root_package_path = "<unknown>"
+        spec = importlib.util.find_spec(package_name)
+        if spec and spec.origin:
+            root_package_path = str(Path(spec.origin).parent)
```
florentLutz, line 392: Could you add a small comment about what that does ? And explain it to me at the same time ? 😅

florentLutz (approved):

PR #641, opened 2025-07-24
Better handling of failed module logging

florentLutz:
## Summary of Changes
This PR slightly alters the handling of the logging of failed modules.

## Related Issues
 For some specific cases were modules were defined and registered through a `RegisterOpenMDAOSystem.explore_folder`, the failed variable was returned as a `set`. this was then passed to the `_log_failed_modules` method which expects `dict` resulting in an Error when doing `failed_modules.items()`. This can be reproduced by adding the disc_4 package in src\fastoad\module_management\tests\data\module_sellar_example and running 
```console
fastoad list_modules src\fastoad\module_management\tests\data\module_sellar_example
```

## New Dependencies
No new dependency

## Checklist

## Additional Context
image

### DIFF
--- a/src/fastoad/module_management/_bundle_loader.py
+++ b/src/fastoad/module_management/_bundle_loader.py
@@ -120,7 +120,20 @@ class BundleLoader:
         if is_package:
             bundles, failed = self._install_python_package(folder_path)
         else:
-            bundles, failed = self.framework.install_package(as_path(folder_path).as_posix(), True)
+            bundles, unformatted_failed = self.framework.install_package(
+                as_path(folder_path).as_posix(), True
+            )
+            # For some failure, the failed object can be returned as a set rather than a dict which
+            # will prevent proper logging of the failures.
+            failed = {}
+            for failed_package in unformatted_failed:
+                path_to_failed_package = (
+                    as_path(folder_path).as_posix()
+                    + "/"
+                    + "/".join(failed_package.split(".")[1:])
+                    + ".py"
+                )
+                failed[failed_package] = path_to_failed_package
 
         for bundle in bundles:
             _LOGGER.info(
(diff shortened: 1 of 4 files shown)


src/fastoad/module_management/_bundle_loader.py
```diff
@@ -128,7 +128,21 @@ def explore_folder(
...
+                reformatted_failed = {}
+                for failed_package in failed:
+                    path_to_failed_package = (
+                        as_path(folder_path).as_posix()
+                        + "/"
+                        + "/".join(failed_package.split(".")[1:])
+                        + ".py"
+                    )
+                    reformatted_failed[failed_package] = path_to_failed_package
+                self._log_failed_modules(reformatted_failed)
```
enricostragiotti, line 143: I would do that after `bundles, failed = self.framework.install_package(as_path(folder_path).as_posix(), True)` and I'll remove the `if isinstance(failed, set):`. Effectively in all of our testing I never encountered a case in which `is_package=False`.

enricostragiotti: But otherwise I'm ok with the logic. As we can't modify directly Pelix I think this is the best we can do

enricostragiotti (changes requested):

src/fastoad/module_management/tests/data/module_sellar_example/disc_4/__init__.py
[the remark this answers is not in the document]
florentLutz: Thanks for the catch, corrected in b44f4866. I can't seem to set up PyCharm so that he uses the proper header for the proper project ^^'

src/fastoad/module_management/tests/data/module_sellar_example/disc_4/disc4.py
```diff
@@ -0,0 +1,42 @@
+"""Sellar discipline 2"""
+#  This file is part of FAST-OAD : A framework for rapid Overall Aircraft Design
+#  Copyright (C) 2024 ONERA & ISAE-SUPAERO
```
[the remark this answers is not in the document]
florentLutz: Thanks again, corrected in b44f4866

src/fastoad/module_management/_bundle_loader.py
florentLutz: Thanks for the suggestion. I moved it were you suggested.
The test I added represents a case where is_package is False.

enricostragiotti (approved):

PR #642, opened 2025-07-28
Better handling of optional module

florentLutz:
Added the FastBundleLoaderUnavailableFactoryError error to handle cases where an id exists but can't be used. It can be used to give the user a custom error message as to what is required for this module.

## Summary of Changes

This module improves the handling of optional modules. The reason for this addition is that some module used by FAST-OAD user can require package or database access which unfortunately can't be made public. Consequently, when handling the imports (which will thus be missing), FAST-OAD will raise a `FastBundleLoaderUnknownFactoryNameError` saying the module is not recognized. However, this particular error is raised regardless of what the actual problem is so there is no way for developers of expliciting the reason why this module can't be used (a private import is missing).

This module introduces the `FastBundleLoaderUnavailableFactoryError` for when a module is unavailable because of a certain optional dependency and when the id should still be recognized.

This is how it's meant to be used: 

```python
import openmdao.api as om

from fastoad.module_management.exceptions import FastBundleLoaderUnknownFactoryNameError

try:
    import optional_dependency
    OPTIONAL_DEPENDENCY = True
except ImportError
    OPTIONAL_DEPENDENCY = False

...

@RegisterOpenMDAOSystem("fastoad.optional_module")
class OptionalModule(om.ExplicitComponent):

        def __init__(self, **kwargs):
            super().__init__(**kwargs)
            
            if not OPTIONAL_DEPENDENCY 
                raise FastBundleLoaderUnavailableFactoryError(
                    "This module is only available if optional_dependency is installed"
                )

```

Trying to use this module in a configuration would previously raise a `FastBundleLoaderUnknownFactoryNameError` but should now raise a FastBundleLoaderUnavailableFactoryError with the following message `This module is only available if optional_module is installed`. Additionally, this module will appear when running 

```console
fastoad list_modules
```

Finally this PR proposes to adopt the difference between unknown/unrecognized module and unavailable module (recognized but can't be used for reasons expected by the developer)

## Checklist

- [X] Have you followed the guidelines in our [Contributing](https://github.com/fast-aircraft-design/FAST-OAD/wiki/Development-environment) wiki?
- [X] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
- [X] Have you added new tests to cover your changes?
- [X] Do the existing tests pass with your changes?
- [X] Have you commented on hard-to-understand areas of the code?

### DIFF
--- a/src/fastoad/module_management/_bundle_loader.py
+++ b/src/fastoad/module_management/_bundle_loader.py
@@ -308,6 +308,7 @@ class BundleLoader:
         :param properties: Initial properties of the component instance
         :return: the component instance
         :raise FastBundleLoaderUnknownFactoryNameError: unknown factory name
+        :raise FastBundleLoaderUnavailableFactoryError: unavailable factory
         """
         with use_ipopo(self.context) as ipopo:
             try:
--- a/src/fastoad/module_management/tests/data/module_sellar_example/disc3/disc3.py
+++ b/src/fastoad/module_management/tests/data/module_sellar_example/disc3/disc3.py
@@ -32,20 +32,8 @@ class RegisteredDisc3(om.ExplicitComponent):
         )
 
     def setup(self):
-        self.add_input("y2", val=1.0)
-
-        self.add_output("y3", val=2.0)
-
-    def setup_partials(self):
-        self.declare_partials("*", "*", method="fd")
+        pass
 
     # pylint: disable=invalid-name
     def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
-        """
-        Evaluates the equation
-        y3 = y2 ** 0.42
-        """
-
-        y2 = inputs["y2"]
-
-        outputs["y3"] = y2**0.42
+        pass
--- a/src/fastoad/module_management/tests/test_register_submodel.py
+++ b/src/fastoad/module_management/tests/test_register_submodel.py
@@ -89,6 +89,12 @@ def test_get_submodel_deactivation(load):
 
 
 def test_unavailable_submodel(load):
+    """
+    Tests the mechanism for unavailable submodel. Unavailable submodel are submodels that are
+    recognized but not expected to be usable unless certain conditions are met, for instance they
+    won't work unless some optional dependencies are installed. It should raise a different error
+    than when the submodel is not recognized.
+    """
     # This model exists AND is available
     RegisterSubmodel.active_models["requirement.2"] = "req.2.submodel.B"
     _ = RegisterSubmodel.get_submodel("requirement.2")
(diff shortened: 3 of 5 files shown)


src/fastoad/module_management/_bundle_loader.py
```diff
@@ -314,6 +315,9 @@ def instantiate_component(self, factory_name: str, properties: dict = None) -> A
                     factory_name, self._get_instance_name(factory_name), properties
                 )
             except TypeError as exc:
+                context = exc.__context__
+                if isinstance(context, FastBundleLoaderUnavailableFactoryError):
+                    raise context from exc
```
enricostragiotti, line 320: So if this is a new possible raised error, it should be in the docstring of the method as well :)

src/fastoad/module_management/tests/data/module_sellar_example/disc3/disc3.py
```diff
@@ -0,0 +1,51 @@
...
+    # pylint: disable=invalid-name
+    def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
+        """
+        Evaluates the equation
+        y3 = y2 ** 0.42
+        """
+
+        y2 = inputs["y2"]
+
+        outputs["y3"] = y2**0.42
```
enricostragiotti, line 51: there is need for all of that?

enricostragiotti: genuine question, I don't have an answer 😅

src/fastoad/module_management/tests/test_register_submodel.py
```diff
@@ -85,3 +86,17 @@ def test_get_submodel_deactivation(load):
 
     with pytest.raises(FastTooManySubmodelsError):
         _ = RegisterSubmodel.get_submodel("requirement.2")
+
+
+def test_unavailable_submodel(load):
```
enricostragiotti, line 91: Can you add here a small description of the test? like what is the objective? it is something that really helps when you aren't familiar with the domain.. Thanks!

src/fastoad/module_management/_bundle_loader.py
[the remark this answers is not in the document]
florentLutz: Thanks :)

src/fastoad/module_management/tests/data/module_sellar_example/disc3/disc3.py
florentLutz: You were right, it is not needed, I also removed the useless bits in disc4.py

src/fastoad/module_management/_bundle_loader.py
florentLutz: Done in 4a63db8 Thanks for the catch !

enricostragiotti (approved):
LGTM! Thanks!

PR #649, opened 2025-08-19
Added compatibility with Python 3.13 v2 (with poetry 2.x)

enricostragiotti:
This PR brings compatibility with Python 3.13.


enricostragiotti:
@esnguyenvan you are definitely right ! I forgot that sorry! If only a tool could look for that....... 😉
Ah, I saw that I commented also the github action caching mechanism. I reworked it a little because before it was not working..

.github/workflows/tests.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     if: ${{ !contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]') }}
     strategy:
       matrix:
-        python-version: [ "3.9", "3.10", "3.12" ]
+        python-version: [ "3.9", "3.11", "3.13" ]
```
florentLutz: Since we are bound to progressively phase out of Python 3.9, I'd suggest testing on 3.10, 3.11 and 3.13. What do you think ?

tests/memory_tests/test_multiple_runs.py
```diff
@@ -68,28 +81,132 @@ def run_problem():
...
+            )
+
+        # Cross-run memory leak detection
+        print("\nMemory leak analysis:")
+        positive_diffs = [
+            diff for _, diff in memory_measurements if diff > 0.1
+        ]  # Only significant positive diffs
+        avg_increase = sum(positive_diffs) / len(positive_diffs)
+        print(f"  Average memory increase per run: {avg_increase:.3f} MiB")
+        if avg_increase > MEMORY_DIFF_THRESHOLD * 0.5:
```
florentLutz: So this is in case the increase is big but not big enough to cause a crash ? For debug then ?

.github/workflows/tests.yml
enricostragiotti: Not yet.. The tests shall pass on the lowes  available python version until we do remove support of python 3.9 entirely

tests/memory_tests/test_multiple_runs.py
```diff
@@ -68,28 +81,132 @@ def run_problem():
...
-    for stat in stats[:10]:
-        print(stat)
-
-    assert final_memory < 5
-    tracemalloc.stop()
+    print_memory_state("After problem outputs")
+
+    # Explicit cleanup attempts
+    try:
+        problem.cleanup()  # If available in OpenMDAO
```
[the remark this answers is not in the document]
enricostragiotti: I got it completely wrong, nice catch !

tests/memory_tests/test_multiple_runs.py
```diff
@@ -68,28 +81,132 @@ def run_problem():
...
+        print("\nMemory measurements per run:")
+        for run_num, diff in memory_measurements:
+            status = "PASS" if abs(diff) <= MEMORY_DIFF_THRESHOLD else "FAIL"
+            print(f"  Run {run_num}: {diff:+.3f} MiB [{status}]")
+
+        print(f"\nOverall memory increase (after garbage collection): {final_diff:.3f} MiB")
+
+        # Overall memory analysis from start to finish
+        print("\nOverall top memory changes (start to finish):")
+        final_snapshot = tracemalloc.take_snapshot()
```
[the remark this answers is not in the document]
enricostragiotti: one is a memory value, a float, the other is a snapshot object

PR #661, opened 2025-11-18
Python 3.14

enricostragiotti:
## Summary of Changes
I didn’t stop at 3.13 :)

In this case everything worked as intended; only a few packages needed their upper bounds updated because of compatibility issues with the new Python version. I also took the opportunity to simplify the dependencies again.

While playing with the dependencies, I updated fastoad-cs25 (which we use for integration tests) and noticed that with the latest version 0.8.0, all the integration tests failed. This was because we had many NaNs in the inputs due to the new variables added by the new loads discipline introduced in this release of cs25. So I regenerated the outputs for the CERAS cases in cs25 and then updated them in the tests.


tests/integration_tests/oad_process/data/CeRAS01_legacy_mission_result.xml
```diff
@@ -1,913 +1,1143 @@
-<!--
```
esnguyenvan: I think that these lines should be kept (just with the year modified)

enricostragiotti: Sorry, my bad, I just copy pasted from CS25 :)

pyproject.toml
```diff
@@ -40,17 +40,14 @@ dependencies = [
     "ipysheet>=0.5.0,<1",
     "plotly>=6.2.0,<7",
     "wop>=2.2.0,<3",
-    "lxml>=4.6.5,<6; python_version<'3.11'",
-    "lxml>=4.9.3,<6; python_version>='3.11'",
+    "lxml>=4.9.3,<7",
```
florentLutz: Doesn't this restrict the number of usable version of lxml ? Since we still support 3.10 shouldn't it be kept ? Same reasoning with other dependencies "simplification" you made.

tests/integration_tests/oad_process/data/CeRAS01_legacy_mission_result.xml
enricostragiotti: Done in 999c3304398a537ba1d143b2105a1e0de06d00f6

esnguyenvan (approved):
Good for me

Issue #637, opened 2025-07-08
Clean up excessive warnings in unit and integration test runs


enricostragiotti:
**Describe the bug**

Currently, running the unit and integration tests for FAST-OAD produces **more than 700 warnings**. While some of these may be harmless or expected (e.g., due to deprecations or third-party packages), the sheer volume:

- Makes it hard to spot real issues
- Slows down development and CI pipelines

**To Reproduce**
Running `pytest`

**Expected behavior**
Let's aim to bring this down to a manageable and meaningful number, ideally close to zero.

PR #667, opened 2026-02-02
Add explicit "unitless" as unit of meausre in registered components

enricostragiotti:
This pull request standardizes the use of units for all relevant OpenMDAO component inputs and outputs, ensuring that variables which are dimensionless explicitly declare `units="unitless"`. This improves model clarity, consistency, and compatibility with OpenMDAO's unit checking and conversion features.

This PR start to reduce the number of warnings found in https://github.com/fast-aircraft-design/FAST-OAD/issues/637

Additionally, we enforce that we NEVER promote a None units of measure. 

PS after this PR, the variables called in the mission file default to 'unitless' if it is not a known parameter (id if not part of the FlightPoint parameters). The following will work:
```yaml
        target:
          altitude: data:mission:sizing:main_route:cruise:altitude_input
```

although this is preferred:
```yaml
        target:
          altitude: 
            value: data:mission:sizing:main_route:cruise:altitude_input
            units: m
```

but the following will not work:
```yaml
        segment: cruise
        time_step: data:mission:sizing:cruise:time_step
```

instead explicitly give the unit: 
```yaml
        segment: cruise
        time_step:
          value: data:mission:sizing:cruise:time_step
          unit: s
```

All the logic of the PR is in src/fastoad/models/performances/mission/mission_definition/mission_builder/input_definition.py and src/fastoad/openmdao/problem.py

PS integration tests would need the unitless changes in cs25 (https://github.com/fast-aircraft-design/FAST-OAD_CS25/tree/copilot/enforce-unitless-connections). I checked locally with the dev branch of cs25 and they pass.


esnguyenvan:
Modified [input_definition.py](https://github.com/fast-aircraft-design/FAST-OAD/pull/667/changes#diff-e4f94a2194cde3b08f9b66ce8d66f88d69011010f51cca5d1ff528b25d9ae6f4) so that the mission declares its inputs with 'untiless' as default when units are not specified in the mission file or if variable in FlightPoint is declared with "-" as unit. Now this PR will work together with https://github.com/fast-aircraft-design/FAST-OAD_CS25/pull/81

src/fastoad/openmdao/problem.py
```diff
@@ -296,19 +298,34 @@ class AutoUnitsDefaultGroup(om.Group):
...
+            if (
+                explicit_units and explicit_units[0] == "n/a"
+            ):  # OpenMDAO returns units="n/a" for discrete variables
+                self.set_input_defaults(name, units=None)
+            elif explicit_units:
+                si_units = [  # If we can find a SI unit among the declared ones, we prefer it.
+                    unit
+                    for unit in explicit_units
+                    if np.isclose(conversion_to_base_units(unit)[0], 0.0)
+                    and np.isclose(conversion_to_base_units(unit)[1], 1.0)
```
esnguyenvan: OK it seems to work and will harmonise the units in xml file

esnguyenvan (approved):
It seems to do the job

esnguyenvan (approved):
OK so now we have:
-explicit unit specification when the parameter is not part of FlightPoint otherwise default to unitless
-SI unit favoured for writing the output xml
I think I am OK with this.
I've update the description of the PR

PR #685, opened 2026-03-23
Add altitude change with regulated thrust

esnguyenvan:
## Summary of Changes

The legacy altitude change segment takes as input the thrust rate and the speed. The slope angle is a result of available thrust and L/D ratio.

This PR reworks the altitude change segment to allow the definition of a prescribed slope angle during altitude change. The thrust rate hence becomes a result of prescribed slope angle and L/D ratio.

A new segment, the regulated_altitude_change is now available and offers the same functionalities as the altitude_change segment (climb to optimal altitude / flight level / CL limit etc...)

## Related Issues

None

## New Dependencies

No new dependencies

## Checklist

- [x] Have you made corresponding changes to the documentation? If yes, please consider providing a link to the updated documentation.
- [x] Does this PR introduce a breaking change? It doesn't

## Additional Context

To be able to keep the same functionalities for both altitude_change and regulated_altitude_change, a new base class BaseAltitudeChange has been created that gathers all common functionalities and inherits only from AbstractLiftFromWeightSegment.

Then, the regulated_altitude_change inherits both from BaseAltitudeChange and from the AbstractRegulatedThrust, that now defines the slope angle and the keyword to control the behaviour in case thrust becomes negative or higher than the maximum thrust.

Finally, altitude change now inherits from BaseAltitudeChange and AbstractManualThrust.


src/fastoad/models/performances/mission/segments/registered/tests/test_altitude_change.py
```diff
@@ -242,7 +242,6 @@ def test_climb_target_CL_at_fixed_mach(polar):
         polar=polar,
         thrust_rate=1.0,
         time_step=2.0,
-        maximum_CL=0.1,
```
enricostragiotti, line 245: well spotted!

src/fastoad/models/performances/mission/segments/registered/cruise.py
```diff
@@ -51,6 +51,10 @@ class CruiseSegment(AbstractRegulatedThrustSegment, AbstractLiftFromWeightSegmen
     the initial value.
     """
 
+    time_step: float = 60
```
enricostragiotti: 60.0

src/fastoad/models/performances/mission/segments/registered/cruise.py
```diff
@@ -51,6 +51,10 @@ class CruiseSegment(AbstractRegulatedThrustSegment, AbstractLiftFromWeightSegmen
...
+
+    slope_angle: float = 0
```
enricostragiotti: 0.0

src/fastoad/models/performances/mission/segments/registered/cruise.py
```diff
@@ -62,6 +66,9 @@ def __post_init__(self):
         ]:
             self.target.mach = AbstractTimeStepFlightSegment.constant_value_name
 
+        # In cruise segment, the slope angle is forced to 0
+        self._slope_angle = 0
```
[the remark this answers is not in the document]
enricostragiotti: yes!

src/fastoad/models/performances/mission/segments/time_step_base.py
```diff
@@ -397,22 +397,31 @@ def compute_propulsion(self, flight_point: FlightPoint):
...
     """
 
-    time_step: float = 60.0
+    # In case the slope angle is forced (climb/descent at constant slope angle)
+    # For cruise, it is zero
+    slope_angle: float = 0.0  # in radian
+
+    # Optional behaviour if the thrust rate is out of limits <0 or >1
+    # extrapolate: means the thrust may be higher of lower than one
+    # limit: means  0 <= thrust_rate <= 1 is forced, when thrust_rate is out of bound,
```
enricostragiotti: to change after the last commit you did, now the bounds are personalized

src/fastoad/models/performances/mission/segments/time_step_base.py
```diff
@@ -397,22 +397,31 @@ def compute_propulsion(self, flight_point: FlightPoint):
...
+    # it switches to manual thrust segment
+    thrust_rate_out_of_bound: str = "extrapolate"
```
enricostragiotti: shouldn't the default behavior be limit?

src/fastoad/models/performances/mission/segments/time_step_base.py
```diff
@@ -397,22 +397,31 @@ def compute_propulsion(self, flight_point: FlightPoint):
...
     def compute_propulsion(self, flight_point: FlightPoint):
-        flight_point.thrust = flight_point.drag
+        flight_point.slope_angle = self.slope_angle
+        flight_point.thrust = flight_point.drag + flight_point.mass * g * np.sin(self.slope_angle)
         flight_point.thrust_is_regulated = True
         self.propulsion.compute_flight_points(flight_point)
 
     def get_gamma_and_acceleration(self, flight_point: FlightPoint) -> tuple[float, float]:
-        return 0.0, 0.0
+        return self.slope_angle, 0.0
```
enricostragiotti, line 426:
```suggestion
        return flight_point.slope_angle, 0.0
```

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+            return flight_points
+
+        if self.thrust_rate_out_of_bound == "extrapolate":
+            # Do nothing, output the flightpoints with thrust rate out of bounds
+            return flight_points
+
+        # Raise a Value error
+        raise ValueError(
+            "The value of option 'thrust_rate_out_of_bound' in regulated_altitude_change is invalid"
+            " it must be one of ['extrapolate', 'limit']"
```
enricostragiotti: slightly hard to maintain, ideally the list ['extrapolate', 'limit'] should be an enum that we can call so that if we add another behavior we don't have to add it here as well. but if it becomes too much of a burden to change it, do not worry and leave it like this

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -30,15 +31,15 @@
...
 )
 from fastoad.models.performances.mission.util import get_closest_flight_level
 
 _LOGGER = logging.getLogger(__name__)  # Logger for this module
 
 
-@RegisterSegment("altitude_change")
 @dataclass
-class AltitudeChangeSegment(AbstractManualThrustSegment, AbstractLiftFromWeightSegment):
+class BaseAltitudeChange(AbstractLiftFromWeightSegment):
```
enricostragiotti, line 43: Shouldn't we change here the fact that if we have a slope angle the lift is not only given by weight but also by thrusts sin(alpha)? AbstractLiftFromWeightSegment assumes that the load factor is 0, but I think we should add also the thrust contribution if the slope angle is different from 0.

I'm interested to know what do you think

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -91,7 +92,7 @@ class AltitudeChangeSegment(AbstractManualThrustSegment, AbstractLiftFromWeightS
     #: with max lift/drag ratio.
     OPTIMAL_FLIGHT_LEVEL = "optimal_flight_level"  # used as constant
 
-    def compute_from_start_to_target(self, start: FlightPoint, target: FlightPoint) -> pd.DataFrame:
+    def _handle_target_settings(self, target, start):
```
enricostragiotti: i would leave the typing helper, they are not changing

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
enricostragiotti: this will break change ALL the missions that use a climb however 😪

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
             target.altitude = optimal_altitude
         else:  # self._original_target_altitude == self.OPTIMAL_FLIGHT_LEVEL:
             target.altitude = get_closest_flight_level(optimal_altitude, up_direction=False)
+
+
+@RegisterSegment("altitude_change")
+@dataclass
+class AltitudeChangeSegment(BaseAltitudeChange, AbstractManualThrustSegment):
```
enricostragiotti, line 221: can you add at least a one liner description for this class? So that the devs knows what to expect, thanks!

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+
+    def compute_from_start_to_target(self, start: FlightPoint, target: FlightPoint) -> pd.DataFrame:
+        self._handle_target_settings(target, start)
+
+        return super().compute_from_start_to_target(start, target)
+
+
+@RegisterSegment("regulated_altitude_change")
+@dataclass
+class RegulatedAltitudeChangeSegment(BaseAltitudeChange, AbstractRegulatedThrustSegment):
```
enricostragiotti, line 244: Same here, please add a short description

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+
+        # Compute the segment with no limitation on thrust_rate
+        flight_points = super().compute_from_start_to_target(start, target)
+
+        # Adjust according to the desired behaviour
+        if self.thrust_rate_out_of_bound == "limit":
+            # Check for out of bound thrust rate and switch to manual thrust segment instead
+
+            if np.any(flight_points.thrust_rate > self.upper_thrust_rate_limit):
+                # We have a too high thrust rate, likely a climb phase, thrust rate forced to 1
```
enricostragiotti:
```suggestion
                # We have a thrust rate too high, likely a climb phase, thrust rate forced to self.upper_thrust_rate_limit
```

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+
+                # We use the last FlightPoint where thrust rate is < 1.0 as a starting point.
```
enricostragiotti: same here

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+                idx = np.argwhere(flight_points.thrust_rate > self.upper_thrust_rate_limit)
+
+                if idx[0] == 0:
+                    # Handle first point of segment being already at thrust rate > 1
```
enricostragiotti: and here

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+                    flight_points.drop(
+                        flight_points.loc[
+                            flight_points.thrust_rate > self.upper_thrust_rate_limit
+                        ].index,
+                        inplace=True,
+                    )
+
+                start.thrust_rate_is_regulated = False
+
+                climb_segment = AltitudeChangeSegment(
```
enricostragiotti, line 304: This is quite a big deviation from what the user asked in the mission file. Wouldn't it make sense for you to add a log here? at least at level info or even warning

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+@dataclass
+class AltitudeChangeSegment(BaseAltitudeChange, AbstractManualThrustSegment):
+    def get_gamma_and_acceleration(self, flight_point: FlightPoint) -> tuple[float, float]:
+        gamma = (flight_point.thrust - flight_point.drag) / flight_point.mass / g
+        return gamma, 0.0
+
+    def compute_from_start_to_target(self, start: FlightPoint, target: FlightPoint) -> pd.DataFrame:
+        self._handle_target_settings(target, start)
+
+        return super().compute_from_start_to_target(start, target)
```
enricostragiotti, line 239:
```suggestion
        return self._handle_target_settings(target, start)
```

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+    # The thrust rate limitations can be changed by the user
+    upper_thrust_rate_limit: float = 1.0
+    lower_thrust_rate_limit: float = 0.0
+
+    def compute_from_start_to_target(self, start: FlightPoint, target: FlightPoint) -> pd.DataFrame:
+
+        self._handle_target_settings(target, start)
+
+        # Compute the segment with no limitation on thrust_rate
+        flight_points = super().compute_from_start_to_target(start, target)
```
enricostragiotti, line 273:
```suggestion
        # Compute the segment with no limitation on thrust_rate
        flight_points = self._handle_target_settings(target, start)
```

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -222,3 +219,121 @@ def _manage_optimal_altitude(self, current, start, target):
...
+                )
+
+                non_regulated_climb_points = climb_segment.compute_from(start)
+                flight_points = pd.concat([flight_points, non_regulated_climb_points]).reset_index(
+                    drop=True
+                )
+
+            return flight_points
+
+        if self.thrust_rate_out_of_bound == "extrapolate":
```
enricostragiotti:
```suggestion
        elif self.thrust_rate_out_of_bound == "extrapolate":
```

docs/documentation/mission_module/mission_file/segments.rst
```diff
@@ -251,6 +251,53 @@ than the original altitude target (the climb stops when the maximum CL is reache
 
     Python documentation: :class:`~fastoad.models.performances.mission.segments.registered.altitude_change.AltitudeChangeSegment`
 
+.. _segment-regulated-altitude-change:
+
+:code:`regulated-altitude_change` segment
```
enricostragiotti: Ok so I'm ok with this name, but  can you think of another usecase of a regulated thrust climb segment other than for the imposed slope with a given speed? Maybe a rate of climb (m/min)? this is the only other thing I can think of. because if there is no other usecase, then I would rename it to something like altitude_change_given_slope, so taht is easier for the users to call

So an additional question would be, what happens if we do not specify a slope angle with this segment? Should we force the user to give this parameter (we should check how to do in this case)?

docs/documentation/mission_module/mission_file/segments.rst
```diff
@@ -251,6 +251,53 @@ than the original altitude target (the climb stops when the maximum CL is reache
...
+        value: 10000.
+        unit: ft
+      equivalent_airspeed: constant
+
+.. code-block:: yaml
+
+    segment: regulated_altitude_change
+    polar: data:aerodynamics:aircraft:cruise    # High speed aerodynamic polar
+    engine_setting: idle
+    slope_angle: -0.1                           # In radian, negative for descent
```
enricostragiotti, line 293:
```suggestion
    slope_angle_deg: -0.1                           # In deg, negative for descent
```

this maybe will reduce the possibility of making mistakes, because i think we can't give units there.. no?

docs/documentation/mission_module/mission_file/segments.rst
```diff
@@ -251,6 +251,53 @@ than the original altitude target (the climb stops when the maximum CL is reache
...
+
+.. list-attributes-for:: regulated-altitude_change
+
+Offers the same functionalities as the altitude change segment, only here the thrust is regulated and the slope angle is prescribed.
+
+.. important::
+
+    **To be used, this segment requires that the propulsion module is able to handle thrust rate > 1 or thrust rate < 0.**
+
+If :code:`thrust_rate_out_of_bound` is set to `limit`, the thrust rate is limited to: 0 <= thrust_rate <= 1.0
```
enricostragiotti: to update

src/fastoad/models/performances/mission/segments/registered/tests/test_regulated_altitude_change.py
```diff
@@ -0,0 +1,399 @@
...
+        last_point = flight_points.iloc[-1]
+        # Note: reference values are obtained by running the process with 2s as time step
+        assert_allclose(last_point.altitude, 5000.0)
+        assert_allclose(last_point.true_airspeed, 150.0)
+
+        assert_allclose(last_point.thrust, 100000 * 0.9, atol=1)
+
+        # Check the gradient is lower than asked at the end of trajectory and that
+        # thrust rate is never>1
+        assert not np.any(flight_points.thrust_rate > 0.9)
```
enricostragiotti: inconsistency

src/fastoad/models/performances/mission/segments/registered/tests/test_regulated_altitude_change.py
```diff
@@ -0,0 +1,399 @@
...
+
+    with pytest.raises(ValueError) as exc_info:
+        segment.compute_from(FlightPoint(altitude=5000.0, mach=0.82, mass=70000.0))
+
+    assert str(exc_info.value) == expected_msg
+
+
+# ------------------------------------------------------------------------------------
+# The following tests are here to ensure that we do not lose the common functionalities
+# of altitude change
```
enricostragiotti, line 374: top! good idea!

src/fastoad/models/performances/mission/segments/registered/tests/test_regulated_altitude_change.py
```diff
@@ -0,0 +1,399 @@
...
+    # therefore the thrust rate must be capped at 1.0.
+    assert_allclose(flight_points.thrust_rate, 1.0, rtol=1e-6)
+
+
+def test_regulated_altitude_change_with_CL_limitation(polar):
+    """Flight is limited by a target lift coefficient (CL)."""
+    propulsion = FuelEngineSet(DummyEngine(5.0e4, 1.0e-5), 2)
+
+    segment = RegulatedAltitudeChangeSegment(
+        target=FlightPoint(CL=0.4418, mach="constant"),
```
enricostragiotti, line 479: so this do not permit the slope to be 0.1? can we test that?

enricostragiotti: just to be sure that the cl max is stronger than the slope angle as constraint

enricostragiotti (changes requested):
Thanks for the PR! This segment will be super useful!
Some minor rework to do, the logic is solid ;)

We should only understand if we want to reuse the regulated thrust alt change for something else, otherwise I would reconsider the naming of the segment..

src/fastoad/models/performances/mission/segments/time_step_base.py
esnguyenvan: No I wouldn't do that, we really want to write the value of the slope angle stored in the segment into the flight_point.slope_angle field, which at this moment is still empty anyway.

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
esnguyenvan: The sin(alpha) component of the thrust is actually always present, in climbing or in cruise segment. But it is usually small as transport aircraft are flying at rather small alpha. We do take it into account in the take off rotation though.

What you are referring to is maybe more the fact that the lift = mg/cos(gamma) and for large values of gamma, this impacts the lift.

Yes we could take it into account for regulated altitude change, since gamma is known in advance, we could modify [compute_lift()](https://github.com/fast-aircraft-design/FAST-OAD/blob/7ba05c76be674a6eef7b102df559299671841948/src/fastoad/models/performances/mission/segments/time_step_base.py#L442) easily. But for manual thrust segment, gamma is calculated at the end, because we first need the drag, that we get from the lift, and if we need gamma to compute the lift... it loops and it's convenient to use small angle approximation to break this loop.

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
esnguyenvan: I'm thinking more of an info rather than a warning as the mission already prints a lot of information. But I agree it should be logged

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
esnguyenvan: Good catch!

It actually was not my initial motivation, it seems that I'm not done working on the function handle_target_settings(). As its name may suggest, it is not supposed to compute the segment. I'm put that in order.

src/fastoad/models/performances/mission/segments/registered/tests/test_regulated_altitude_change.py
esnguyenvan: It's not stronger, the slope angle is not a target, it's a field. So the aircraft will just climb at a constant slope angle until the target CL.

docs/documentation/mission_module/mission_file/segments.rst
esnguyenvan: I think it is better to use radians for the trajectory, because it's equivalent to giving the slope in % and if someone wants to target a certain climb rate, the angle is just climb_rate / aircraft_true_airspeed.

docs/documentation/mission_module/mission_file/segments.rst
esnguyenvan: A certain climb rate would translate into the slope angle as: climb_rate/aircraft_true_airspeed. So an altitude change with a cst climb_rate can be done if the option true_airspeed=constant is set.

It would require more work to make it available with EAS/CAS cst, maybe something for later development ?

Although the philosophy remains the same, I'm not sure we need another segment for regulated altitude change with cst climb rate. For example, it could be another field within the same segment

As of what happens if we do not set the slope angle... well it's default value is 0 so it won't reach the target. Maybe I change the default value to None so that we can raise a value error in case the user forgets to set it.

src/fastoad/models/performances/mission/segments/time_step_base.py
esnguyenvan: As you wish

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
esnguyenvan: You're right about the maintainability, I've implemented it.

src/fastoad/models/performances/mission/segments/time_step_base.py
esnguyenvan: Implement 'limit' as the default behaviour

docs/documentation/mission_module/mission_file/segments.rst
esnguyenvan: UPdated

esnguyenvan:
I've should have addressed all the comments. I've actually made quite some changes due to the fact that self._handle_target_settings(target, start) was evaluating the segment, which was not the intended behaviour. Now it is fixed.

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
enricostragiotti: We discussed that before, thanks for the explanation!

esnguyenvan:
@enricostragiotti , I've fix the bad behaviour you identified the other day. Now it should be all fine.

src/fastoad/models/performances/mission/segments/registered/altitude_change.py
```diff
@@ -121,39 +123,20 @@ def compute_from_start_to_target(self, start: FlightPoint, target: FlightPoint)
...
+        ):
+            # If CL of the starting point is above the max CL, we stop the climb/descent
+            _LOGGER.warning(
+                'The first point in a segment of "%s" has a CL = %.2f > maximum_CL = %.2f. '
+                'Skipping "altitude_change" segment.',
+                self.name,
+                start.CL,
+                self.maximum_CL,
+            )
+            target.altitude = start.altitude  # to avoid any processing
```
enricostragiotti:
```suggestion
            target = deepcopy(start)  # to avoid any processing
```

enricostragiotti (approved):

PR #689, opened 2026-04-01
Add TOFL based on CS-25.113(a)

enricostragiotti:
This pull request adds support for estimating and outputting the Takeoff Field Length (TOFL) for takeoff phases in the mission performance model. It introduces logic to identify takeoff phases, computes the TOFL as 1.15 times the takeoff distance, and ensures this value is only present when a takeoff phase is done (we only output automatic variables on the xml for phases and not segments).

New tests have been added to validate the correct calculation and presence of the TOFL output.


enricostragiotti:
ready to go @esnguyenvan!

esnguyenvan (approved):
Good for me

PR #690, opened 2026-04-10
Cumulative quantities in FlightPoint

florentLutz:
## Summary of Changes

- Expands the notion of cumulative quantities in the FlightPoint class with a way of defining how they are incremented (it doesn't apply to quantities by default in the FlightPoint)
- Expands time-step-based segments so that cumulative quantities are automatically incremented (will not work for other segments and Breguet-type cruise)
- Adds an example of how this could be used for hybrid propulsion (random example)

## Checklist

- [x] Have you made corresponding changes to the documentation? If yes, please consider providing a link to the updated documentation.

### DIFF
--- a/src/fastoad/model_base/flight_point.py
+++ b/src/fastoad/model_base/flight_point.py
@@ -99,6 +99,35 @@ class FlightPoint:
             # Removing a field, even an original one
             >>> FlightPoint.remove_field("sfc")
 
+    **Time integration**
+
+        For those additional fields, it is also possible to define their time derivative. If a time
+        derivative is defined this way, the mission model will automatically perform the
+        integration. This integration will be achieved by incrementing the field at each time step
+        by the product of the time derivative and the time step in seconds.
+
+        The order in which a field and its time derivative are defined is irrelevant. However,
+        the existence of the field containing the time derivative will be verified when the
+        analysis process is run::
+
+            # Adding a time-dependent field and declaring the field which contains its derivative
+            # using the 'integrates_from' descriptor.
+            >>> FlightPoint.add_field(
+            ...    "electric_energy",
+            ...    unit="J",
+            ...    default_value=0.0,
+            ...    is_cumulative=True,
+            ...    integrates_from="electric_power",
+            ... )
+
+            # Now defining the field that contains the derivative. The unit of the derivative must
+            # be the unit of the original field *per second*
+            >>> FlightPoint.add_field(
+            ... "electric_power", default_value=0.0, unit="W", is_cumulative=False
+            ... )
+
+            # Field will be incremented as: field[i+1] = field[i] + time_derivative[i] * time_step
+
     .. note::
 
         All original parameters in FlightPoint instances are expected to be in SI units.
--- a/src/fastoad/model_base/flight_point.py
+++ b/src/fastoad/model_base/flight_point.py
@@ -101,17 +101,18 @@ class FlightPoint:
 
     **Time integration**
 
-        For those additional fields, it is also possible to define their time derivative. If a time
-        derivative is defined this way, the mission model will automatically perform the
-        integration. This integration will be achieved by incrementing the field at each time step
-        by the product of the time derivative and the length of the time step in seconds.
+        For those additional user-defined fields, it is also possible to define their time
+        derivative. If a time derivative is defined this way, the mission model will
+        automatically perform the integration. This integration will be achieved by incrementing
+        the field at each time step by the product of the time derivative and the length of the
+        time step in seconds.
 
         The order in which a field and its time derivative are defined is irrelevant. However,
         the existence of the field containing the time derivative will be verified when the
         analysis process is run::
 
             # Adding a time-dependent field and declaring the field which contains its derivative
-            # using the 'integrates_from' descriptor.
+            # using the 'integrates_from' field descriptor.
             >>> FlightPoint.add_field(
             ...    "electric_energy",
             ...    unit="J",
@@ -126,7 +127,8 @@ class FlightPoint:
             ...     "electric_power", default_value=0.0, unit="W", is_cumulative=False
             ... )
 
-            # Field will be incremented as: field[i+1] = field[i] + time_derivative[i] * time_step[i]
+            # Field will be incremented as: field[i+1] = field[i] + time_derivative[i] *
+            # time_step[i]
 
     .. note::
 
@@ -453,11 +455,10 @@ class FlightPoint:
                         # integrable. Worst case, the process won't crash but the results will be
                         # inconsistent
                         if not cls_field.metadata[FIELD_DESCRIPTOR].is_cumulative:
-                            _LOGGER.warning(
-                                "Field '%s' is declared as integrating from another field but is "
-                                "not declared as cumulative. 'integrates_from' will only be used if"
-                                " 'is_cumulative' is True.",
-                                field_name,
+                            raise ValueError(
+                                f"Field '{field_name}' is declared as integrating from another "
+                                f"field but is not declared as cumulative. 'integrates_from' can "
+                                f"only be declared if 'is_cumulative' is True.",
                             )
 
                         # Check that the field we want to integrate from actually exists. If it
@@ -474,7 +475,7 @@ class FlightPoint:
         return cls.__time_integrable_quantities
 
     @classmethod
-    def get_time_integrand(cls, field_name) -> str | None:
+    def get_time_integrand(cls, field_name) -> str:
         """
         Returns the quantity the field integrates from.
         """
--- a/src/fastoad/model_base/flight_point.py
+++ b/src/fastoad/model_base/flight_point.py
@@ -15,6 +15,7 @@
 
 from __future__ import annotations
 
+import logging
 from collections.abc import Mapping, Sequence
 from copy import deepcopy
 from dataclasses import asdict, dataclass, field, fields
@@ -27,6 +28,8 @@ from fastoad._utils.arrays import scalarize
 from fastoad.constants import EngineSetting
 
 FIELD_DESCRIPTOR = "field_descriptor"
+
+_LOGGER = logging.getLogger(__name__)  # Logger for this module
 
 
 @dataclass
@@ -200,7 +203,7 @@ class FlightPoint:
 
     # Will store field metadata when needed. Must be accessed through _get_field_descriptors()
     __field_descriptors: ClassVar[dict] = {}
-    __cumulative_quantities: ClassVar[dict] = {}
+    __time_integrable_quantities: ClassVar[dict] = {}
 
     def __post_init__(self):
         self._relative_parameters = {"ground_distance", "time"}
@@ -346,7 +349,7 @@ class FlightPoint:
                      physical quantity. Set to None, when unit concept does not apply.
         :param is_cumulative: True if field value is summed up during mission
         :param integrates_from: Name of the other field this value is integrated by (in other words,
-                                incremented by that value multiplied by the time step)
+                                incremented by that value multiplied by the time step always in s)
         """
         cls._redeclare_fields()
 
@@ -404,27 +407,49 @@ class FlightPoint:
         return cls._get_field_descriptors().get(field_name, _FieldDescriptor(None, None))
 
     @classmethod
-    def get_cumulative_quantities(cls) -> dict[str, str]:
+    def get_time_integrable_quantities(cls) -> dict[str, str | None]:
         """
         Uses this method instead of accessing cls.__cumulative_quantity to ensure it
         will always be correctly populated.
         """
-        if not cls.__cumulative_quantities:
-            cls.__cumulative_quantities = {
-                cls_field.name: cls_field.metadata[FIELD_DESCRIPTOR].integrates_from
-                for cls_field in fields(cls)
-                if not cls_field.name.startswith("_")
-                and cls_field.metadata[FIELD_DESCRIPTOR].is_cumulative
-            }
+        if not cls.__time_integrable_quantities:
+            for cls_field in fields(cls):
+                field_name = cls_field.name
+                if not field_name.startswith("_"):
+                    time_derivative = cls_field.metadata[FIELD_DESCRIPTOR].integrates_from
+                    if time_derivative is not None:
+                        # Field is declared as having a derivative through the integrated_from
+                        # descriptor but not through the is_cumulative descriptor. It is not
+                        # consistent, so we raise a warning and ignore the field as being
+                        # integrable. Worst case, the process won't crash but the results will be
+                        # inconsistent
+                        if not cls_field.metadata[FIELD_DESCRIPTOR].is_cumulative:
+                            _LOGGER.warning(
+                                "Field '%s' is declared as integrating from another field but is "
+                                "not declared as cumulative. 'integrates_from' will only be used if"
+                                " 'is_cumulative' is True.",
+                                field_name,
+                            )
 
-        return cls.__cumulative_quantities
+                        # Check that the field we want to integrate from actually exists. If it
+                        # doesn't raise an error because the process will actually crash.
+                        if time_derivative not in cls._get_field_descriptors():
+                            raise ValueError(
+                                f"Field '{field_name}' is declared as integrating from "
+                                f"'{time_derivative}', but '{time_derivative}' is not an existing "
+                                f"field.",
+                            )
+
+                        cls.__time_integrable_quantities[field_name] = time_derivative
+
+        return cls.__time_integrable_quantities
 
     @classmethod
-    def get_cumulative_quantity(cls, field_name) -> str | None:
+    def get_time_integrand(cls, field_name) -> str | None:
         """
-        Returns the quantity the field cumulates from if it is cumulative and if it was specified.
+        Returns the quantity the field integrates from.
         """
-        return cls.get_cumulative_quantities().get(field_name, None)
+        return cls.get_time_integrable_quantities().get(field_name, None)
 
     @classmethod
     def _redeclare_fields(cls):
@@ -440,4 +465,4 @@ class FlightPoint:
             )
 
         cls.__field_descriptors = {}  # Will need to rebuild this dict on next usage.
-        cls.__cumulative_quantities = {}  # Will need to rebuild this dict on next usage.
+        cls.__time_integrable_quantities = {}  # Will need to rebuild this dict on next usage.
--- a/src/fastoad/model_base/flight_point.py
+++ b/src/fastoad/model_base/flight_point.py
@@ -104,7 +104,7 @@ class FlightPoint:
         For those additional fields, it is also possible to define their time derivative. If a time
         derivative is defined this way, the mission model will automatically perform the
         integration. This integration will be achieved by incrementing the field at each time step
-        by the product of the time derivative and the time step in seconds.
+        by the product of the time derivative and the length of the time step in seconds.
 
         The order in which a field and its time derivative are defined is irrelevant. However,
         the existence of the field containing the time derivative will be verified when the
@@ -123,7 +123,7 @@ class FlightPoint:
             # Now defining the field that contains the derivative. The unit of the derivative must
             # be the unit of the original field *per second*
             >>> FlightPoint.add_field(
-            ... "electric_power", default_value=0.0, unit="W", is_cumulative=False
+            ...     "electric_power", default_value=0.0, unit="W", is_cumulative=False
             ... )
 
             # Field will be incremented as: field[i+1] = field[i] + time_derivative[i] * time_step
--- a/src/fastoad/model_base/flight_point.py
+++ b/src/fastoad/model_base/flight_point.py
@@ -36,8 +36,8 @@ class _FieldDescriptor:
     """
 
     is_cumulative: bool | None = False
-    cumulates_from: str | None = (
-        None  # Indicates the field in the flight point that increments that values
+    integrates_from: str | None = (
+        None  # Indicates the field in the flight point that is used to integrate that value
     )
     unit: str | None = None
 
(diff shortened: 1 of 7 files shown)


florentLutz:
The more I think about it, the more I think a distinction should be made in the names (methods and variables) between cumulative quantities and something that has to do with time derivatives

florentLutz:
@esnguyenvan @enricostragiotti Bear with me, it's my first PR in a while, and with Copilot 😅 I'm going to apply its suggestions and re-request you for the review of the PR 😄

src/fastoad/model_base/flight_point.py
```diff
@@ -394,6 +403,29 @@ def _get_field_descriptor(cls, field_name) -> _FieldDescriptor:
...
+                cls_field.name: cls_field.metadata[FIELD_DESCRIPTOR].integrates_from
+                for cls_field in fields(cls)
+                if not cls_field.name.startswith("_")
+                and cls_field.metadata[FIELD_DESCRIPTOR].is_cumulative
+            }
+
+        return cls.__cumulative_quantities
+
+    @classmethod
+    def get_cumulative_quantity(cls, field_name) -> str | None:
```
esnguyenvan: If I'm not mistaken, this function is used only during tests. Although it could be used somewhere else as a utility function, the main work is done by FlightPoint.get_cumulative_quantities()

Both method have very close name, it's prone to confusion. I would rename this function to avoid confusing an important method and a utility method. Here are two suggestions:
```suggestion
    def get_integrated_from(cls, field_name) -> str | None:
```

```suggestion
    def get_cumulated_from(cls, field_name) -> str | None:
```

src/fastoad/model_base/tests/test_flight_point.py
```diff
@@ -109,3 +109,31 @@ def test_descriptors():
...
+    # cumulates from
+
+    FlightPoint.add_field(
+        "foo", annotation_type=float, default_value=42.0, unit="m", is_cumulative=False
+    )
+    FlightPoint.add_field(
+        "bar",
+        annotation_type=float,
+        default_value=1337.0,
+        unit="slug/ft",
```
esnguyenvan: Really !

tests/dummy_plugins/dist_2/dummy_plugin_2/models/subpackage/dummy_engine.py
```diff
@@ -61,3 +61,54 @@ def get_model(self, inputs) -> IPropulsion:
...
+        component.add_input("data:propulsion:dummy_engine:electric_efficiency", 0.95, units="unitless")
+        component.add_input("data:geometry:propulsion:engine_count", 2)
+
+    def get_model(self, inputs) -> IPropulsion:
+        return FuelEngineSet(
+            DummyHybridEngine(
+                inputs["data:propulsion:dummy_engine:max_thrust"],
+                inputs["data:propulsion:dummy_engine:hybridization_ratio"],
+                inputs["data:propulsion:dummy_engine:max_sfc"],
+                inputs["data:propulsion:dummy_engine:electric_efficiency"],
```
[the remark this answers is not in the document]
esnguyenvan: I think we don't care about that suggestion

src/fastoad/models/performances/mission/tests/test_mission.py
```diff
@@ -203,3 +203,90 @@ def test_mission(low_speed_polar, high_speed_polar, propulsion):
...
+    # Check that the electric energy consumed since the start has been constructed properly
+    assert_allclose(
+        (
+            flight_points.electric_energy.iloc[1:101].to_numpy()
+            - flight_points.electric_energy.iloc[:100].to_numpy()
+        )
+        / (flight_points.time.iloc[1:101].to_numpy() - flight_points.time.iloc[:100].to_numpy()),
+        flight_points.electric_power.iloc[:100].to_numpy(),
+        1e-6,
+    )
```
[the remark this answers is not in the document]
esnguyenvan: I think we don't care about that suggestion

src/fastoad/model_base/flight_point.py
```diff
@@ -340,6 +345,8 @@ def add_field(
         :param unit: expected unit for the added field. "-" should be provided for a dimensionless
                      physical quantity. Set to None, when unit concept does not apply.
         :param is_cumulative: True if field value is summed up during mission
+        :param integrates_from: Name of the other field this value is integrated by (in other words,
+                                incremented by that value multiplied by the time step)
         """
         cls._redeclare_fields()
```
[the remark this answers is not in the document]
esnguyenvan: I would apply this suggestion or the one in time_step_base.py

src/fastoad/model_base/tests/test_flight_point.py
```diff
@@ -109,3 +109,31 @@ def test_descriptors():
...
+        is_cumulative=True,
+        integrates_from="foo",
+    )
+
+    assert FlightPoint.is_cumulative("bar")
+    assert "bar" in FlightPoint.get_cumulative_quantities()
+    assert FlightPoint.get_cumulative_quantity("bar") == "foo"
```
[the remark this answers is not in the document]
esnguyenvan: I think we don't care about that

src/fastoad/models/performances/mission/tests/test_mission.py
esnguyenvan: Could we check the final value of electric_power as well? Would help identifying any problem happening throughout the different segments.

esnguyenvan (changes requested):
Hey! Great job!
That will considerably increase the flexibility of the mission, it remains simple and well integrated into the mission algorithm.

I have minor suggestions. I would as well update the documentations as this is an important new feature.

tests/dummy_plugins/dist_2/dummy_plugin_2/models/subpackage/dummy_engine.py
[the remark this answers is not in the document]
florentLutz: Actually yes but no. He is right but the wrapper alltogether is unused so I'll just go ahead and delete it 😄

src/fastoad/models/performances/mission/segments/time_step_base.py
```diff
@@ -377,6 +378,21 @@ def distance_to_optimum(altitude):
...
+    def _increment_cumulative_quantities(
+        next_point: FlightPoint, previous_point: FlightPoint, time_step
+    ):
+        for cumulative_field, integrand_field in FlightPoint.get_cumulative_quantities().items():
+            if integrand_field:  # To avoid incrementing default cumulative quantity
+                setattr(
+                    next_point,
+                    cumulative_field,
+                    getattr(previous_point, cumulative_field)
+                    + getattr(previous_point, integrand_field) * time_step,
```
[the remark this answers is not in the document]
florentLutz: Following what you said in the other comment, I closed @esnguyenvan. By construction, we now ensure that the field we try to integrate from now actually exists. The worst thing that can happen now is that the propulsion model does not fill in the value of the integrand field correctly or that it takes the default value, which, for some reason, was fixed at a non-numeric value. Therefore, I think it is the responsibility of the developer of the propulsion model to make sure it is handled correctly. We could add a check here, but given that it is a line of code that is executed countless times in the mission, it would quickly be computationally expensive.

src/fastoad/model_base/flight_point.py
florentLutz: Before you made the suggestion, I updated this function's name to `get_time_integrand` to make it clearer that we are dealing with time integrals. I did the same to several other functions. What do you think about it?

src/fastoad/model_base/tests/test_flight_point.py
florentLutz: What can I say, I really wanted to celebrate the birth of Robert III of Scotland.

If you were referring to the units, thanks for pointing out that the time derivative didn't have the right units, it is now in "slug/s" instead of "m" 🐌

src/fastoad/models/performances/mission/tests/test_mission.py
florentLutz: Done, thanks for the advice 😄

florentLutz:
Thanks for the patience @enricostragiotti, I've applied Copilot's suggestion and updated the documentation (available at https://fast-oad.readthedocs.io/en/cumulative_quantity_flight_points/) and docstring of the flightPoint class. I've also applied your suggestions @esnguyenvan.

src/fastoad/model_base/flight_point.py
esnguyenvan: Sorry, I didn't see your message before adding my review. I'm on it

src/fastoad/model_base/flight_point.py
```diff
@@ -394,6 +435,51 @@ def _get_field_descriptor(cls, field_name) -> _FieldDescriptor:
         """
         return cls._get_field_descriptors().get(field_name, _FieldDescriptor(None, None))
 
+    @classmethod
+    def get_time_integrable_quantities(cls) -> dict[str, str | None]:
```
esnguyenvan, line 441: Nice! Much clearer indeed

src/fastoad/model_base/flight_point.py
```diff
@@ -394,6 +435,51 @@ def _get_field_descriptor(cls, field_name) -> _FieldDescriptor:
...
+                                f"'{time_derivative}', but '{time_derivative}' is not an existing "
+                                f"field.",
+                            )
+
+                        cls.__time_integrable_quantities[field_name] = time_derivative
+
+        return cls.__time_integrable_quantities
+
+    @classmethod
+    def get_time_integrand(cls, field_name) -> str | None:
```
esnguyenvan: Perfect

src/fastoad/models/performances/mission/segments/time_step_base.py
```diff
@@ -377,6 +378,24 @@ def distance_to_optimum(altitude):
             ).root
         )
 
+    @staticmethod
+    def _increment_cumulative_quantities(
+        next_point: FlightPoint, previous_point: FlightPoint, time_step
+    ):
+        for (
+            integrable_field,
```
esnguyenvan, line 386: Much clearer as well!

src/fastoad/models/performances/mission/segments/time_step_base.py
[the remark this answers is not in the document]
esnguyenvan: Since the cumulative and the integrands fields will most likely always be declared and manipulated by the propulsion module, I think it's reasonable.

esnguyenvan (approved):
Yup, I approve the new names!

src/fastoad/model_base/tests/test_flight_point.py
```diff
@@ -109,3 +109,66 @@ def test_descriptors():
...
+        integrates_from="foo",
+    )
+    with caplog.at_level("WARNING"):
+        FlightPoint.get_time_integrable_quantities()
+
+    assert (
+        "Field 'bar' is declared as integrating from another field but is not declared as "
+        "cumulative. 'integrates_from' will only be used if 'is_cumulative' is True."
+        in caplog.records[0].message
+    )
```
enricostragiotti: I think that in this case an error would be preferable, no? it is a TypeError or Value error for me.

In my head the two possible behaviuous are:

1. error, something like 

```python
if integrates_from is not None and not is_cumulative:
    raise ValueError(
        f"Field '{name}' declares 'integrates_from' but is not cumulative."
    )
```

2. warning, but on the top of that you add
```python
 if integrates_from is not None:
    is_cumulative = True
```

because otherwise you throw a warning but what's next?

src/fastoad/model_base/flight_point.py
```diff
@@ -93,6 +99,35 @@ class FlightPoint:
             # Removing a field, even an original one
             >>> FlightPoint.remove_field("sfc")
 
+    **Time integration**
+
+        For those additional fields, it is also possible to define their time derivative. If a time
```
enricostragiotti: `those `is unclear to me here

src/fastoad/model_base/flight_point.py
```diff
@@ -93,6 +99,35 @@ class FlightPoint:
...
+        derivative is defined this way, the mission model will automatically perform the
+        integration. This integration will be achieved by incrementing the field at each time step
+        by the product of the time derivative and the length of the time step in seconds.
+
+        The order in which a field and its time derivative are defined is irrelevant. However,
+        the existence of the field containing the time derivative will be verified when the
+        analysis process is run::
+
+            # Adding a time-dependent field and declaring the field which contains its derivative
+            # using the 'integrates_from' descriptor.
```
enricostragiotti: descriptor?
https://docs.python.org/3/glossary.html#term-descriptor

those are dataclass fields no? Or eventually class parameters

src/fastoad/model_base/flight_point.py
```diff
@@ -197,6 +232,7 @@ class FlightPoint:
 
     # Will store field metadata when needed. Must be accessed through _get_field_descriptors()
     __field_descriptors: ClassVar[dict] = {}
+    __time_integrable_quantities: ClassVar[dict] = {}
```
enricostragiotti, line 237: Do you know why a double __ is used here?

src/fastoad/model_base/flight_point.py
```diff
@@ -394,6 +435,51 @@ def _get_field_descriptor(cls, field_name) -> _FieldDescriptor:
...
+                        # consistent, so we raise a warning and ignore the field as being
+                        # integrable. Worst case, the process won't crash but the results will be
+                        # inconsistent
+                        if not cls_field.metadata[FIELD_DESCRIPTOR].is_cumulative:
+                            _LOGGER.warning(
+                                "Field '%s' is declared as integrating from another field but is "
+                                "not declared as cumulative. 'integrates_from' will only be used if"
+                                " 'is_cumulative' is True.",
+                                field_name,
+                            )
```
enricostragiotti, line 462: as said, i don't agree with the way this is handled here. see my comments before

src/fastoad/model_base/flight_point.py
```diff
@@ -394,6 +435,51 @@ def _get_field_descriptor(cls, field_name) -> _FieldDescriptor:
...
+
+                        cls.__time_integrable_quantities[field_name] = time_derivative
+
+        return cls.__time_integrable_quantities
+
+    @classmethod
+    def get_time_integrand(cls, field_name) -> str | None:
+        """
+        Returns the quantity the field integrates from.
```
enricostragiotti, line 480:
```suggestion
        Returns the quantity the field integrates from. Return None if...
```

tests/dummy_plugins/dist_2/dummy_plugin_2/models/subpackage/dummy_engine.py
```diff
@@ -61,3 +61,34 @@ def get_model(self, inputs) -> IPropulsion:
...
+
+        :param max_thrust: thrust when thrust rate = 1.0
+        :param hybridization_ratio: share of thermal power, between 0.0 and 1.0
+        :param max_sfc: SFC when thrust rate = 1.0
+        :param eta_electric: electric branch efficiency
+        """
+        self.max_thrust = max_thrust
+        self.hybridization_ratio = hybridization_ratio
+        self.max_sfc = max_sfc
+        self.eta_electric = eta_electric
```
enricostragiotti, line 85: you just recoded a dataclass ;)

src/fastoad/models/performances/mission/tests/test_mission.py
```diff
@@ -203,3 +203,96 @@ def test_mission(low_speed_polar, high_speed_polar, propulsion):
...
+    FlightPoint.add_field(
+        "electric_energy",
+        annotation_type=float,
+        default_value=0.0,
+        unit="J",  # SI unit for energy is Joules, even though W*h are more convenient.
+        # Besides, the integrand will be multiplied by the time step in s
+        is_cumulative=True,
+        integrates_from="electric_power",
+    )
```
enricostragiotti: Here you are adding two class variables, but you are never releasing them. So if in the same pytest call we would have access to these vars even after the end of this specific unit test. I would wrap everithing like this:
```python
FlightPoint.remove_field("electric_energy")
FlightPoint.remove_field("electric_power")

try:
        add the test here
finally:
        FlightPoint.remove_field("electric_energy")
        FlightPoint.remove_field("electric_power")
```

src/fastoad/models/performances/mission/tests/test_mission.py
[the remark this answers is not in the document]
enricostragiotti: I just saw this now... yes we should take that into account!

src/fastoad/model_base/flight_point.py
```diff
@@ -93,6 +99,35 @@ class FlightPoint:
...
+            ...    integrates_from="electric_power",
+            ... )
+
+            # Now defining the field that contains the derivative. The unit of the derivative must
+            # be the unit of the original field *per second*
+            >>> FlightPoint.add_field(
+            ...     "electric_power", default_value=0.0, unit="W", is_cumulative=False
+            ... )
+
+            # Field will be incremented as: field[i+1] = field[i] + time_derivative[i] * time_step
```
enricostragiotti:
```suggestion
            # Field will be incremented as: field[i+1] = field[i] + time_derivative[i] * time_step[i]
```

enricostragiotti (changes requested):
Some small changes needed, but the PR is effective, clear, easy to understand. Thanks for that!

src/fastoad/model_base/flight_point.py
florentLutz: Isn't that what they are called?

```python
@dataclass
class _FieldDescriptor:
    """
    Class to be used as dataclass field metadata.
    """

    is_cumulative: bool | None = False
    integrates_from: str | None = (
        None  # Indicates the field in the flight point that is used to integrate that value
    )
    unit: str | None = None
```

src/fastoad/model_base/flight_point.py
florentLutz: Nope 😅 I trust the previous coder who developed that and since the functionalities were similar, I just used the same format

src/fastoad/model_base/flight_point.py
florentLutz: Actually, it should never return `None`.  Using `get_time_integrable_quantities()` ensure that anything that is in `__time_integrable_quantities` has a valid entry. So it will never return None. I'll update the function signature.

tests/dummy_plugins/dist_2/dummy_plugin_2/models/subpackage/dummy_engine.py
florentLutz: I just adapted the other dummy engine 😅 We could change both if you really think there is an added value to it

src/fastoad/model_base/tests/test_flight_point.py
florentLutz: Thanks for the suggestions. Out of what you proposed, I prefer to go with a straight-up `ValueError`.

src/fastoad/model_base/flight_point.py
florentLutz: In the context of the docstring, it seemed pretty clear to me as the previous paragraph explained how to add an additional field, but I've clarified it (I hope)

src/fastoad/models/performances/mission/tests/test_mission.py
florentLutz: I went with a simple solution, hope it works for you. Also took the opportunity to clean up the other tests where I've added fields and not removed them

src/fastoad/model_base/flight_point.py
enricostragiotti: FYI https://www.reddit.com/r/learnpython/comments/1seba3/what_is_name_mangling/

src/fastoad/model_base/flight_point.py
enricostragiotti: I think Christophe called them like this erroneously, those for me are just fields.. But it is not important, it is just that descriptors are a reserved word in python that is definitely not that

src/fastoad/models/performances/mission/tests/test_mission.py
```diff
@@ -296,3 +296,6 @@ def test_hybrid_mission(low_speed_polar, high_speed_polar, hybrid_propulsion):
         flight_points.electric_power.iloc[:100].to_numpy(),
         1e-6,
     )
+
+    FlightPoint.remove_field("electric_energy")
+    FlightPoint.remove_field("electric_power")
```
enricostragiotti: using try finally is usually a better idea just because the resources are freed even in the case of an error being thrown in your code that does not let you reach those resource freeing lines

tests/dummy_plugins/dist_2/dummy_plugin_2/models/subpackage/dummy_engine.py
enricostragiotti: nono no need

src/fastoad/model_base/flight_point.py
enricostragiotti: perfect thanks!

src/fastoad/model_base/tests/test_flight_point.py
enricostragiotti: LGTM!

enricostragiotti (changes requested):
LGTM I would just add a small comment on the test on the need of the `FlightPoint.remove_field("....")` lines. If you do not want to use the try except is fine for me even if I don't see the point in not doing that :)

florentLutz:
@enricostragiotti And what would that comment be? 😅 
Regarging the `try ... except ... finally`, I have no fundamental issue against it I just wanted to be consistent with what was previously done. Do you want me to change it ?

enricostragiotti:
> @enricostragiotti And what would that comment be? 😅 Regarging the `try ... except ... finally`, I have no fundamental issue against it I just wanted to be consistent with what was previously done. Do you want me to change it ?

For me the comment would just be something like # In this test we are modifying class variables, so we need to check that they are freed at the end of the test in order to not influence following tests.

this is just for us for remembering why there is the need to do that. And yes as is not a problem for you I would use try finally

enricostragiotti:
Thanks @florentLutz everything is green for me!

enricostragiotti (approved):

PR #697, opened 2026-06-10
Addition of option to avoid copying an output file in local folder with Calc_runner

### LINKED ISSUE #600: Avoid Copying Output Files in CalcRunner Folders
**Describe the bug**
Currently, when launching calculations with CalcRunner, a separate folder is created for each DOE point. In this folder, CalcRunner correctly copies the configuration file, mission file, and input file, with necessary path modifications, allowing the calculation to be relaunched directly from the folder.

However, the issue is that output files are also copied into the folder at the start, which causes several problems:
- It creates confusion by mixing input and output files that do no not belong to the same point.
- It prevents us from determining whether a calculation has converged since the output files are already present before execution.

Suggested Fix:
- Prevent output files from being copied at the start, allowing them to be generated naturally when the calculation runs.
- Optionally, introduce a status file or marker to track whether the calculation has been executed and converged.

rparello:
Currently when using Calc_runner, a local folder is created in which all the required files to run FAST-OAD on the specific point are copied. This includes the output file if the output file specified in the configuration file already exists. This is problematic because if the FAST-OAD run on this point fails, an output file will still be present in the local folder, making the detection of failed runs complex.  

## Summary of Changes

- An option was added to _make_path_local to not copy the file in the local folder, default behavior is unchanged and will copy the file
- When running Calc_runner, the option is used to not copy the output file in the local calc folder

## Related Issues

closes #600

### DIFF
--- a/src/fastoad/io/configuration/configuration.py
+++ b/src/fastoad/io/configuration/configuration.py
@@ -300,7 +300,7 @@ class FASTOADProblemConfigurator:
 
         self._data[KEY_INPUT_FILE] = self._make_path_local(self.input_file_path, new_folder_path)
         self._data[KEY_OUTPUT_FILE] = self._make_path_local(
-            self.output_file_path, new_folder_path, copy=copy_output_file
+            self.output_file_path, new_folder_path, copy_path=copy_output_file
         )
         if copy_models:
             new_model_folders = []
@@ -420,7 +420,7 @@ class FASTOADProblemConfigurator:
         original_path: str | PathLike,
         new_folder_path: str | PathLike,
         local_path: str | PathLike | None = None,
-        copy: bool | int = True,  # noqa: FBT001, FBT002
+        copy_path: bool | int = True,  # noqa: FBT001, FBT002
     ) -> str:
         """
         For 'original_path' "/foo/bar/baz[.ext]", returns "./baz[.ext]" or
@@ -432,7 +432,7 @@ class FASTOADProblemConfigurator:
         :param original_path:
         :param new_folder_path:
         :param local_path:
-        :param copy: Whether to copy the original file or folder. If False, only the path is
+        :param copy_path: Whether to copy the original file or folder. If False, only the path is
         modified, but no file is copied.
         :return: the relative path
         """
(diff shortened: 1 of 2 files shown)


enricostragiotti (approved):

enricostragiotti (approved):
Thanks! That was one that stayed a lot in the back of my mind ;)

src/fastoad/io/configuration/configuration.py
```diff
@@ -411,6 +420,7 @@ def _make_path_local(
         original_path: str | PathLike,
         new_folder_path: str | PathLike,
         local_path: str | PathLike | None = None,
+        copy: bool | int = True,  # noqa: FBT001, FBT002
```
florentLutz: I'd choose a more explicit name for the argument, especially since copy is also the name of a package. Something like `copy_path` perhaps

src/fastoad/io/configuration/configuration.py
florentLutz, line 419: As a side note on this function, which isn't directly related to the PR, but is the docstring is clear enough ? To me, from the docstring, it appears that this function only returns a "localized" version of the original path. It actually does more than that by copying files. Is it just me ?

florentLutz (commented):
Thanks you for the PR, as usual I'm very picky on some details but you can merge anyway if you'd like

src/fastoad/io/configuration/configuration.py
rparello: We hesitated about changing the name internally too, but I didn't know it was the name of a package. I applied your suggestion in [e09e260](https://github.com/fast-aircraft-design/FAST-OAD/pull/697/commits/e09e2607dafb94683b761a89136e9e92e3a7c2a5)

src/fastoad/io/configuration/configuration.py
rparello: If you read the full docstring, it does talk about  copying the file or folder, so I don't really see the problem

src/fastoad/io/configuration/configuration.py
florentLutz: Actually, it's more of a builtin module than a package.
Anyway, thanks for the change :)

src/fastoad/io/configuration/configuration.py
florentLutz: Yes it does, and its just me being picky but it says so but not at the beginning more as a side note. Any there is no problem, it does not require any action as I said in my original message

PR #700, opened 2026-06-30
Addition of a way to pass options to propulsion wrapper

### LINKED ISSUE #699: PropulsionWrapper should be able to declare options
I'm always frustrated when (😛) I want to create a Propulsion model which might use different levels of fidelity for different models of the powertrain. But as Propulsion model are not components, model_options cannot be used and the mission doesn't pass down it option to the propulsion

Mission could define a generic option dictionary that users could use to pass down option to the propulsion model using model options. For instance:
```python
    self.options.declare(
            "propulsion_option",
            default={},
            types=dict,
            desc="Option dictionnary for propulsion model",
        )
```
Then, in the .yml, users could use `model_options` to write some generic options for the powertrain, for instance

```yml
model_options:
  "*":
    propulsion_option: {"propeller_fidelity": "ADT"}
```
And the, in the setup of the propulsion wrapper

```python
@RegisterPropulsion("multi_fidelity_propulsion_wrapper")
class MultiFidelityPropulsionWrapper(IOMPropulsionWrapper):

    def setup(self, component: Component):
        if component.options["propulsion_option"]["propeller_fidelity"] == "ADT":
            ...
        else:
            ....
```

florentLutz:
## Summary of Changes

This PR adds a way to pass options to the propulsion wrappers used in the computation of the aircraft's performance. Also adds a test to showcase how it can be used.

## Related Issues
This PR closes #699 

## Checklist

- [x] Have you made corresponding changes to the documentation? Yes: https://fast-oad.readthedocs.io/en/add_propulsion_options/documentation/custom_modules/add_propulsion_module.html

### DIFF
--- a/docs/documentation/custom_modules/add_propulsion_module.rst
+++ b/docs/documentation/custom_modules/add_propulsion_module.rst
@@ -160,6 +160,7 @@ Passing options to the wrapper
 
 When developing your propulsion model, you might want to pass options to change how the model behaves and declares inputs. You can do that by accessing the `propulsion_options` option inside the wrapper::
 
+    from openmdao.core.component import Component
     import fastoad.api as oad
 
 
--- a/docs/documentation/custom_modules/add_propulsion_module.rst
+++ b/docs/documentation/custom_modules/add_propulsion_module.rst
@@ -154,3 +154,42 @@ with the `propulsion_id` keyword, as in following example:
       performance:
         id: fastoad.performances.mission
         propulsion_id: star.trek.propulsion
+
+Passing options to the wrapper
+==============================
+
+When developing your propulsion model, you might want to pass options to change how the model behaves and declares inputs. You can do that by accessing the `propulsion_options` option inside the wrapper::
+
+    import fastoad.api as oad
+
+
+    @oad.RegisterPropulsion("star.wars.propulsion")
+    class WarpDriveWrapper(oad.IOMPropulsionWrapper):
+
+        def setup(self, component: Component):
+
+            if component.options["propulsion_options"].get("fidelity_level", None) == "high_fidelity":
+                [ ... ]
+            else:
+                [ ... ]
+
+
+Options can then be declared directly in the configuration file by using :ref:`model options<configuration-model-options>`.
+
+.. code-block:: yaml
+
+    title: OAD Process with custom propulsion model using options
+
+    [ ... ]
+
+    # Definition of OpenMDAO model
+    model:
+      [ ... ]
+      performance:
+        id: fastoad.performances.mission
+        propulsion_id: star.wars.propulsion
+
+    model_options:
+      '*':
+        propulsion_options:
+          fidelity_level: high_fidelity
--- a/docs/documentation/custom_modules/add_propulsion_module.rst
+++ b/docs/documentation/custom_modules/add_propulsion_module.rst
@@ -194,3 +194,20 @@ Options can then be declared directly in the configuration file by using :ref:`m
       '*':
         propulsion_options:
           fidelity_level: high_fidelity
+
+Propulsion options can also be declared by passing them directly to the performance module.
+
+.. code-block:: yaml
+
+    title: OAD Process with custom propulsion model using options
+
+    [ ... ]
+
+    # Definition of OpenMDAO model
+    model:
+      [ ... ]
+      performance:
+        id: fastoad.performances.mission
+        propulsion_id: star.wars.propulsion
+        propulsion_options:
+          fidelity_level: high_fidelity
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -134,6 +134,30 @@ def with_dummy_plugin_2():
 
 
 @pytest.fixture
+def with_dummy_plugin_3():
+    """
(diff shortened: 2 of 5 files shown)


src/conftest.py
```diff
@@ -133,6 +133,30 @@ def with_dummy_plugin_2():
...
+def with_dummy_plugin_3():
+    """
+    Reduces plugin list to dummy-dist-2 with plugin test_plugin_3
+    (two configuration files, model folder, no notebooks, 3 source data files).
+
+    Any previous state of plugins is restored during teardown.
+    """
+    _setup()
+    dummy_dist_2 = Mock(importlib_metadata.Distribution)
+    dummy_dist_2.name = "dummy-dist-2"
```
rparello, line 146: I don't fully grasp the purpose of those lines, but is it normal to have dummy_dist_2 in dummy_plugin_3 ? shouldn't it be 3 everywhere ?

docs/documentation/custom_modules/add_propulsion_module.rst
```diff
@@ -154,3 +154,43 @@ with the `propulsion_id` keyword, as in following example:
...
+        def setup(self, component: Component):
+
+            if component.options["propulsion_options"].get("fidelity_level", None) == "high_fidelity":
+                [ ... ]
+            else:
+                [ ... ]
+
+
+Options can then be declared directly in the configuration file by using :ref:`model options<configuration-model-options>`.
+
```
rparello, line 179: You could also specify the options for propulsion under perfomance in the yml right ? Should we illustrate that too ?

tests/dummy_plugins/dist_2/dummy_plugin_3/models/subpackage/dummy_engine_with_options.py
```diff
@@ -0,0 +1,125 @@
...
+        flight_point.sfc = self.sfc
+
+
+class DummyEngineHighFidelity(AbstractFuelPropulsion):
+    def __init__(self, max_thrust, sfc_min_thrust_rate, sfc_max_thrust_rate):
+        """
+        Dummy engine model.
+
+        Max thrust does not depend on flight conditions.
+        SFC varies linearly between a value when thrust rate is nil and another value when thrust
```
rparello:
```suggestion
        SFC varies linearly between a value when thrust rate is nul and another value when thrust
```

rparello:
Hey, I tried to review the PR but I have trouble understanding how the propulsion wrapper works within the mission (especially how the new options are transmitted to a potential new propulsion module). Aside from the minor comments I made, I may have miss some other important things; a second review may be required before merging.

src/conftest.py
florentLutz: Actually, `dummy_plugin_3` is in `dummy_dist_2` 😅 
There is a difference between a plugin (like `fast-oad-cs25` or `fast-oad-core`) and a distribution, which might include multiple plugins (like `fast-oad`).
For testing, though, you might want specific test plugins to be loaded instead of loading the entire distribution, which is what I am doing here. Loading the `dummy_plugin_3` (by adding it to the entry_points), and telling it where that plugin is located (in that case, the `dummy_dist_2`)

Long story short, it shouldn't be 3 everywhere. I hope my explanation helps.

docs/documentation/custom_modules/add_propulsion_module.rst
florentLutz: You are totally right, I've expanded the documentation in 44e75943

florentLutz:
Hey @rparello 👋 , thanks for the review.

The performance and propulsion models are two of the trickiest parts in the current FAST-OAD implementation, as they are not exactly straight-up OpenMDAO components.

To better understand how the interface works, I recommend checking out the documentation on the interface used: https://fast-oad.readthedocs.io/en/v1.10.0/documentation/custom_modules/add_propulsion_module.html#the-ipropulsion-interface

Regarding how options are passed, the magic happens in the `setup` function that the `IOMPropulsionWrapper` requires. That function is called in the `setup` method of any component that requires the propulsion model to be run (in particular, the mission). And what the `setup` of `IOMPropulsionWrapper` does is add, to the components in which it is called, the inputs required by the propulsion module. It does that by altering the components themselves, which are passed as an argument. For instance, in the rubber engine of the `fast-oad-cs25` plugin:

```python
def setup(self, component: Component):
        component.add_input("data:propulsion:rubber_engine:bypass_ratio", np.nan, units="unitless")
        component.add_input(
            "data:propulsion:rubber_engine:overall_pressure_ratio", np.nan, units="unitless"
        )
        component.add_input(
            "data:propulsion:rubber_engine:turbine_inlet_temperature", np.nan, units="K"
        )
        ...
```

As for any component, it is thus possible to get the options by doing `component.options`.

I hope this helps you understand better ✨

tests/dummy_plugins/dist_2/dummy_plugin_3/models/subpackage/dummy_engine_with_options.py
rparello: i just realized that it should be null in fact, not nul, sorry

PR #703, opened 2026-08-04
Fix XDSM generation

florentLutz:
## Summary of Changes
This PR fixes a bug in the XDSM generation which had gone unnoticed due to some tests being bypassed. After the introduction of new linting rules, `dry_run` became a named argument in the method actually calling WhatsOpt, while in the CLI and API it wasn't. As the usage of the `write_xdsm` method of the `whatsopt.py` script wasn't updated, this led to an error preventing the generation of the XDSM. This can easily be reproduced, pre-fix, using the CLI.

## Checklist

- [x] Does this PR introduce a breaking change? No !

### DIFF
--- a/src/fastoad/cmd/api.py
+++ b/src/fastoad/cmd/api.py
@@ -659,7 +659,9 @@ def write_xdsm(
     problem.setup()
     problem.final_setup()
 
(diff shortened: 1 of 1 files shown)


enricostragiotti (approved):
Nice spot!

PR #705, opened 2026-09-23
Allow OpenMDAO up to 3.45 and lock 3.45.1

ScottDelbecq:
Watchman tests have failed nightly for at least three months because they install the OpenMDAO development version, while `main` was capped below 3.43.

Since OpenMDAO 3.44, the driver keeps a weakref to its problem, and `deepcopy` does not update it. The problem copy made by `ProblemAnalysis` therefore had a driver working on the original, not set up, model (`'NoneType' object has no attribute '_abs_get_val'`). `get_mpi_safe_problem_copy()` now rebinds the driver to the copy.

Since 3.43, dynamic shapes are resolved in `final_setup()`, so two tests in `test_problem.py` no longer read values between `setup()` and `final_setup()`.

The pin is raised to `<3.46`, and 3.45.1 is locked.


enricostragiotti:
Hello @ScottDelbecq, thank you very much for this PR, it was really needed. I started looking into it 2 weeks ago, but I didn't had the time to dig into it.

I do not know if you wanted to continue working on this PR or if it is ready to be rewieved, but as there are quite some other PRs in the making, it would be cool to finally have the watchman test working again. Nice findings anyway, the error message was indeed quite difficult to decrypt!

Ok for merging for me :)

enricostragiotti (approved):

